PR #1, opened 2019-01-29
Initial

sayplastic:


.circleci/config.yml
```diff
@@ -0,0 +1,18 @@
+version: 2
+jobs:
+  release:
+    docker:
+      - image: circleci/golang:1.10
```
lyoshenka: you should use go 1.11 so you can take advantage of [go modules](https://github.com/golang/go/wiki/Modules)

PR #4, opened 2019-02-15
Implement video playback from the reflectors

sayplastic:


sayplastic:
@lyoshenka I would appreciate any comments you might have regarding this (the API endpoint is not there yet but that part is rather trivial)

player/player.go
```diff
@@ -0,0 +1,227 @@
...
+}
+
+func (s *reflectedStream) URL() string {
+	return reflectorURL + string(s.SDHash)
+}
+
+// resolve gets SD hash from the daemon for a given URI to enable further retrieval of stream content from
+// reflector servers
+func (s *reflectedStream) resolve() error {
+	client := ljsonrpc.NewClient(config.Settings.GetString("Lbrynet"))
```
lyoshenka: i recommend passing this value into newReflectedStream() instead of using a global config

player/player.go
```diff
@@ -0,0 +1,227 @@
...
+	response, err := client.Resolve(s.URI)
+	if err != nil {
+		return err
+	}
+	source := (*response)[s.URI].Claim.Value.Stream.Source
```
lyoshenka: what if `s.URI` is not set?

player/player.go
```diff
@@ -0,0 +1,227 @@
...
+	DecodedIV     []byte `json:"-"`
+	Length        int    `json:"length"`
+	encryptedBody []byte
+}
+
+// PlayURI downloads and streams LBRY video content located at uri delimited by rangeHeader
+// (use rangeHeader := request.Header.Get("Range"))
+func PlayURI(uri string, rangeHeader string, writer http.ResponseWriter) (err error) {
+	start, end := parseRange(rangeHeader)
+	stream := newReflectedStream(uri)
```
lyoshenka: this var is shadowing lbry.go/stream include

player/player.go
```diff
@@ -0,0 +1,227 @@
...
+		"url": s.URL(),
+	}).Info("requesting stream data")
+	httpRequest, err := http.NewRequest("GET", s.URL(), new(bytes.Buffer))
+	httpClient := &http.Client{}
+	httpResponse, err := httpClient.Do(httpRequest)
+	if err != nil {
+		return err
+	}
+	defer httpResponse.Body.Close()
+	err = json.NewDecoder(httpResponse.Body).Decode(&s)
```
lyoshenka: I recommend using the types in lbry.go/stream as much as possible. we already have types defined for an SDBlob, and other useful parts.

Instead of doing a lot of the stuff in this function, you read httpResponse.Body into a byte slice, then do

```
sd := &stream.SDBlob{}
sd.FromBlob(data)
```

Now you have a proper sd blob. You can iterate through the BlobInfos and add those up to get the size. Or even better, you could add a function to SDBlob that takes a startBytes and endBytes and returns which blobs those bytes are in.

player/player.go
```diff
@@ -0,0 +1,227 @@
...
+		"blobs_sizes":  blobsSizes,
+		"stream_size":  s.Size,
+		"uri":          s.URI,
+	}).Info("got stream data")
+	return nil
+}
+
+// prepareWriter writes necessary range headers and sets the status, preparing HTTP for partial blob streaming.
+// This should be called before streamBlobs
+func (s *reflectedStream) prepareWriter(startByte, endByte int64, writer http.ResponseWriter) (startBlob, endBlob int) {
```
lyoshenka: this function does two separate things: calculate the byte ranges and write headers. they should not be tied together.

player/player.go
```diff
@@ -0,0 +1,227 @@
...
+			}
+			writer.Write(decryptedBody)
+		} else {
+			return errors.Err("server responded with an unexpected status (%v)", httpResponse.Status)
+		}
+	}
+	return nil
+}
+
+func (b *reflectedBlob) decrypt(s *reflectedStream) (decryptedBody []byte, err error) {
```
lyoshenka: i added https://github.com/lbryio/lbry.go/commit/969a1423827178deed34ee77acfeb1e781f8f18d for you

PR #10, opened 2019-03-06
Update headers

seanyesmunt:


server/server.go
```diff
@@ -42,8 +42,9 @@ func NewConfiguredServer() *Server {
 		InterruptChan:  make(chan os.Signal),
 		DefaultHeaders: make(map[string]string),
 	}
-	s.DefaultHeaders["X-Access-Control-Allow-Origin"] = "*"
-	s.DefaultHeaders["Server"] = "lbry.tv"
+	s.DefaultHeaders["Access-Control-Allow-Origin"] = "*"
+	s.DefaultHeaders["Content-Type"] = "application/json; charset=utf-8"
```
sayplastic: Unfortunately cannot set server-wide content-type to `application/json` — we are also serving video and plain html here. I'll fix it for `/api/proxy` endpoint though.

server/server.go
```diff
@@ -42,8 +42,9 @@ func NewConfiguredServer() *Server {
...
+	s.DefaultHeaders["Server"] = "api.lbry.tv"
```
sayplastic: This is just cosmetic/vanity [HTTP header](https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Server) and I believe it won't affect anything even if we put `youtube.com` there.

server/server.go
seanyesmunt: Ah. Right.

PR #12, opened 2019-03-07
Implement io.ReadSeeker interface for reflectedStream

sayplastic:


Makefile
```diff
@@ -39,4 +39,4 @@ clean:
 
 .PHONY: server
 server:
-	go run . serve
+	LW_DEBUG=1 go run . serve
```
lyoshenka: careful committing this, and all the other debug stuff

player/player.go
```diff
@@ -48,47 +50,76 @@ func PlayURI(uri string, rangeHeader string, w http.ResponseWriter) (err error)
...
+	var zero, newSeekOffset int64
+
+	if whence == io.SeekEnd {
+		newSeekOffset = s.Size - 1 - offset
+	} else if whence == io.SeekStart {
+		newSeekOffset = offset
+	} else if whence == io.SeekCurrent {
+		newSeekOffset = s.seekOffset + offset
+	} else {
+		return zero, e.New("invalid seek whence argument")
```
lyoshenka: why do you need a `zero` var? why not just return `0`?

Makefile
sayplastic: makefile is only for development, where debug mode by default makes sense

player/player.go
sayplastic: I think it was a brain fart

sayplastic:
@lyoshenka I decided to man up and copy/paste/patch streaming http method from `net/http` to make my implementation work. Turns out non-streaming formats delivery also works fine when ranges are implemented correctly. Could you please give it a cursory glance?

PR #26, opened 2019-05-20
Get stream size from source metadata

sayplastic:
## PR Checklist
Please check all that apply to this PR using "x":

- [x] I have checked that this PR is not a duplicate of an existing PR (open, closed or merged)
- [x] I have checked that this PR does not introduce a breaking change
- [x] This PR introduces breaking changes and I have provided a detailed explanation below

## PR Type
What kind of change does this PR introduce?

Why is this change necessary?

- [ ] Bugfix
- [ ] Feature
- [ ] Breaking changes (bugfix or feature that introduces breaking changes)
- [ ] Code style update (formatting)
- [ ] Refactoring (no functional changes)
- [ ] Documentation changes
- [x] Other - Please describe:

Player module should use stream metadata content size if present (or the old method of manual estimation otherwise).

### DIFF
--- a/player/player_test.go
+++ b/player/player_test.go
@@ -15,7 +15,7 @@ const streamURL = "what#6769855a9aa43b67086f9ff3c1a5bacb5698a27a"
 func Test_newReflectedStream(t *testing.T) {
 	rs, err := newReflectedStream(streamURL)
 	if err != nil {
-		t.Fatal(err)
+		panic(err)
 	}
 	assert.Equal(t,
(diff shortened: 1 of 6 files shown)


player/player_test.go
```diff
@@ -15,7 +15,7 @@ const streamURL = "what#6769855a9aa43b67086f9ff3c1a5bacb5698a27a"
 func Test_newReflectedStream(t *testing.T) {
 	rs, err := newReflectedStream(streamURL)
 	if err != nil {
-		t.Fatal(err)
+		panic(err)
```
tiger5226: @sayplastic why panic instead of `t.Fatal` curious of the use case for a test.

PR #28, opened 2019-05-24
Feature/accounts

sayplastic:
This PR brings account support to lbrytv.
Authentication is done through internal-apis.


db/users.go
```diff
@@ -0,0 +1,85 @@
...
+// GetUserByToken retrieves user by internal-api auth_token
+// TODO: Refactor out into logical pieces
+// TODO: Implement different error types for better error messages
+func GetUserByToken(token string) (*models.User, error) {
+	var u *models.User
+	ctx := context.Background()
+
+	u, err := models.Users(models.UserWhere.AuthToken.EQ(token)).OneG(ctx)
+	if err != nil {
+		Conn.Logger.LogF(monitor.F{"token": token}).Info("token supplied but isn't present in the database, trying internal-apis")
```
tiger5226: hmmm, should we really be logging user tokens? This seems like a security issue. I would just log their user id or something.

db/users.go
```diff
@@ -0,0 +1,85 @@
...
+		c := lbryinc.NewClient(token)
+		c.ServerAddress = config.Settings.GetString("InternalAPIHost")
```
tiger5226: I would recommend that you have a const for these sort of hardcoded strings. This is a recipe for bugs a year from now when it gets changed and a search for the string is not performed, especially if there is reuse.

tiger5226: see here https://github.com/lbryio/chainquery/blob/master/config/config.go#L28

db/users.go
tiger5226: btw this is in multiple areas, just referencing it once though.

db/users.go
```diff
@@ -0,0 +1,85 @@
...
+				return u, nil
+			}
+		}
+	}
+	return u, nil
+}
+
+// GetAccountIDFromRequest retrieves SDK  account_id of a user making a http request
+// by a header provided by the http client
+func GetAccountIDFromRequest(req *http.Request) (string, error) {
```
tiger5226: So in our [api](https://github.com/lbryio/lbry.go/tree/master/extras/api) package do a search for `FormValues`. It's actually a really clean way for handling parameters with validators too so you dont have to do the below. We use it extensively in internal-apis. I like it alot. I use it in chainquery as well.

db/users.go
```diff
@@ -0,0 +1,85 @@
...
+				u.Seed = a.Seed
+				err = u.InsertG(ctx, boil.Infer())
+
+				if err != nil {
+					Conn.Logger.LogF(monitor.F{"token": token, "email": email}).Error("error inserting a record, rolling back account")
+					_, errDelete := lbrynet.RemoveAccount(email)
+					if errDelete != nil {
+						Conn.Logger.LogF(monitor.F{"token": token, "email": email}).Errorf("error rolling back account: %v", errDelete)
+					}
+					return nil, err
```
tiger5226: I would recommend that you use the `lbry.go/extras/errors` package. It's a sweet addition grin made that keeps track of the stack trace so an api call can return the stack trace of an error when in debug mode. It is EXTREMELY helpful when debugging an api call. When this goes into full use, you will thank me when @tzarebczan gives you a failing api call as a reproduction. instead of `return err`, you just write ` return errors.Err(err)`, and you only have to do this on the first error not the rollups. The rest is handled by the package.

lbrytv.docker.yml
```diff
@@ -2,5 +2,5 @@ StaticDir: /static
 BaseContentURL: https://api.lbry.tv/content/
 SentryDSN: "https://9ef75ef77a844d4aa35d834d78266477:3cff15dc36fa4595b056af781b042852@sentry.io/1416147"
 DatabaseConnection: postgres://postgres:postgres@postgres
-DatabaseName: lbrytv
+DatabaseName: postgres
```
tiger5226: I think lbrytv is better. Why the change?

proxy/processors.go
```diff
@@ -0,0 +1,116 @@
...
+			resultArray[0]["claim_name"], resultArray[0]["claim_id"],
+			resultArray[0]["file_name"])
+	}
+	response.Result = resultArray
+	return response, nil
+}
+
+func getDefaultAccount(accounts *ljsonrpc.AccountListResponse) *ljsonrpc.Account {
+	for _, account := range accounts.LBCMainnet {
+		fmt.Println(account)
```
tiger5226: looks like leftover from testing.

tiger5226: Also what happens here when there are 100K accounts? Should this just be updated and stored periodically?

sqlboiler.toml
```diff
@@ -0,0 +1,8 @@
+[psql]
+  dbname = "lbrytv"
```
tiger5226: hmmm looks like it was changed above but not here.

db/users.go
```diff
@@ -0,0 +1,85 @@
...
+				}
+
+				u = new(models.User)
+				u.Email = email
+				u.AuthToken = token
+				u.IsIdentityVerified = false
+				u.HasVerifiedEmail = false
+				u.SDKAccountID = a.ID
+				u.PrivateKey = a.PrivateKey
```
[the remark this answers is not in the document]
sayplastic: That's a valid concern. I intentionally didn't implement any security here because I feel this needs to be thought through and not give a false sense of safety just because we store an encrypted value in the database. Plus there's this whole sync functionality that we haven't applied to lbrytv yet.

lbrytv.docker.yml
sayplastic: The motivation behind the change is to keep the defaults. But you're right, a persistent database should be named.

db/users.go
sayplastic: `FormValues` looks good. Seems like it's just for, well, forms. I'm retrieving a header value here.

db/conn.go
```diff
@@ -45,60 +69,65 @@ func GetURL(params connectionParams) string {
...
+}
 
-	db.Exec(fmt.Sprintf("DROP DATABASE %s;", pq.QuoteIdentifier(config.Settings.GetString("DatabaseName"))))
-	if db.Error != nil {
-		panic(db.Error)
-	}
+// CreateDB creates the requested database.
+func CreateDB(dbName string) (err error) {
+	c := NewConnection(GetDSN(ConnParams{DatabaseName: "postgres"}))
+	_, err = c.DB.Exec(fmt.Sprintf("create database %s;", pq.QuoteIdentifier(dbName)))
```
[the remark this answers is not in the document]
sayplastic: I remember we had a a very similar discussion in a different request:) I'm reasonably aware of sql injection dangers but postgresql libs do not handle query placeholders in schema changing queries. I worked around it by feeding a value processed by `pq.QuoteIdentifier`, this should be safe enough.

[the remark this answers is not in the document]
sayplastic: Perhaps I should place a comment there so no-one tries to use this practice for regular queries.

sayplastic:
@tiger5226 really appreciate the time you took to review the code brother and your displeasure with the scope of changes is entirely reasonable, I'll do my best to push out smaller sets of changes in the future.

I [documented the proxy package](https://github.com/lbryio/lbrytv/pull/28/files#diff-3ac192716a1df66bbd1008d327af8792R1), let me know if the documentation is not useful.

db/migrations/0001_initial.sql
```diff
@@ -0,0 +1,22 @@
...
+    "email" varchar NOT NULL,
+    "auth_token" varchar NOT NULL,
+    "is_identity_verified" bool NOT NULL DEFAULT false,
+    "has_verified_email" bool NOT NULL DEFAULT false,
+    "sdk_account_id" varchar NOT NULL,
+    "private_key" varchar NOT NULL,
+    "public_key" varchar NOT NULL,
+    "seed" varchar NOT NULL,
+    PRIMARY KEY ("id"),
+    UNIQUE ("auth_token"),
```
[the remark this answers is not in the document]
sayplastic: @tiger5226 @nikooo777 Is this documented somewhere?

db/migrations/0001_initial.sql
```diff
@@ -0,0 +1,22 @@
+-- +migrate Up
+
+-- +migrate StatementBegin
+CREATE TABLE "users" (
+    "id" serial,
+    "created" timestamp NOT NULL DEFAULT now(),
```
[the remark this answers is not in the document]
sayplastic: I'd rather not rely on SQL hooks for modified at, as not every row update should be tracked.

db/users.go
```diff
@@ -0,0 +1,85 @@
...
+	"github.com/lbryio/lbrytv/lbrynet"
+	"github.com/lbryio/lbrytv/models"
+	"github.com/lbryio/lbrytv/monitor"
+
+	"github.com/lbryio/lbry.go/extras/lbryinc"
+	"github.com/volatiletech/sqlboiler/boil"
+)
+
+// TokenHeader is the name of HTTP header which is supplied by client and should contain internal-api auth_token
+const TokenHeader string = "X-Lbry-Auth-Token"
```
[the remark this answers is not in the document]
sayplastic: HTTP is an encompassing RFC containing the definition of _headers_, hence the use of HTTP. 

Parameters are just as easily intercepted as headers, GET parameters even more so. But lbrytv code has no notion of transport channel, it is wrapped in SSL by a reverse proxy in front of it.

db/conn_test.go
```diff
@@ -0,0 +1,77 @@
...
+	rows, err = c.DB.Query("SELECT id FROM users")
+	require.NotNil(t, err)
+
+	c.MigrateUp()
+	rows, err = c.DB.Query("SELECT id FROM users")
+	require.Nil(t, err)
+	rows.Close()
+
+	c.MigrateDown()
+	rows, err = c.DB.Query("SELECT id FROM users")
```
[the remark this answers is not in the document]
sayplastic: Not possible to close an empty result

db/conn.go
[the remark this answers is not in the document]
tiger5226: yeah, just noting its not a pattern just an exception.

db/migrations/0001_initial.sql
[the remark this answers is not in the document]
tiger5226: sure, makes sense. I dont use them in chainquery but we do use them in internal-apis. Main thing is the `_at`

db/migrations/0001_initial.sql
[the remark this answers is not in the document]
tiger5226: No I don't think so. It's part of the model for internal-apis. I am sure there is an issue somewhere. I think it was about a year ago when I implemented that change. Generally there is no documentation of internal-apis besides the code.

db/users.go
tiger5226: Yeah agreed. I guess I also wasn't sure why you chose headers vs parameters. I assumed it was a trivial decision and was suggesting to use params solely to leverage the FormValues since it makes things so much easier.

sayplastic: Request body contains SDK query already, there’s no space for params. And I’d rather not transfer sensitive data in the URL.

db/token_locks.go
```diff
@@ -0,0 +1,34 @@
...
+	queue int32
+}
+
+var tokenLocker = make(map[string]*authToken)
+
+var logger = monitor.NewModuleLogger("token_locks")
+
+func lockToken(token string) {
+	if _, ok := tokenLocker[token]; !ok {
+		tokenLocker[token] = &authToken{}
```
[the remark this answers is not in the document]
sayplastic: Yeah, fair point and I already ran into it.

db/conn.go
```diff
@@ -5,29 +5,53 @@ import (
 
 	"github.com/lbryio/lbrytv/config"
 	"github.com/lbryio/lbrytv/monitor"
+	"github.com/lib/pq"
+	"github.com/volatiletech/sqlboiler/boil"
 
-	"github.com/jinzhu/gorm"
+	"github.com/gobuffalo/packr/v2"
```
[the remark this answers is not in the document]
sayplastic: Yeah, I'm using the rule you mentioned. This just slipped past me

db/conn.go
```diff
@@ -45,60 +69,67 @@ func GetURL(params connectionParams) string {
...
-	dbURL := GetURL(connectionParams{DatabaseName: "postgres"})
+func connect(dialect, dsn string) (*sqlx.DB, error) {
+	conn, err := sqlx.Connect(dialect, dsn)
+	return conn, err
+}
 
-	if dbName == "" {
-		panic("DatabaseName not configured")
+// MigrateUp executes forward migrations.
+func (c ConnData) MigrateUp() {
```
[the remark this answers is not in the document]
sayplastic: Thought about it for some time and figured that both `ConnData` and `Logger` are shallow structures containing pointers to larger things, i.e. passing them around by value instead of pointer should be alright, as long as we're not modifying them.

db/users.go
sayplastic: I remembered trying it before but couldn't because it [string-serializes to some gibberish in tests](https://github.com/lbryio/lbry.go/issues/62). Will try it again when I figure out a way to fix it.

kauffj:
How close is this?

sayplastic:
@kauffj I cannot merge this because it depends on 0.38 and last week attempt showed it's not ready to be deployed on lbrytv.

PR #56, opened 2019-08-15
Feature/benchmarking

sayplastic:
1. Mask account_id in SDK query logs

2. Improve metrics collector. Documentation for metrics collector is provided in the package source files, hope it is clear enough!


app/proxy/service.go
```diff
@@ -91,12 +91,28 @@ func (q *Query) Params() interface{} {
...
 // ParamsAsMap returns query params converted to plain map.
 func (q *Query) ParamsAsMap() map[string]interface{} {
-	if paramsMap, ok := q.Params().(map[string]interface{}); ok {
+	paramsMap := make(map[string]interface{})
+	if paramsStrings, ok := q.Params().(map[string]string); ok {
+		for k, v := range paramsStrings {
+			paramsMap[k] = v
+		}
+		return paramsMap
+	} else if paramsMap, ok := q.Params().(map[string]interface{}); ok {
```
tiger5226: won't this always be true? if you are returning a `map[string]interface{}` always why would we ever try to do `map[string]string`?

app/proxy/service.go
```diff
@@ -236,12 +252,12 @@ func (c *Caller) call(rawQuery []byte) (*jsonrpc.RPCResponse, CallError) {
 	}
 	execTime := time.Now().Sub(queryStartTime).Seconds()
 
-	c.service.LogExecTime(q.Method(), execTime, q.Params())
+	c.service.SetMetricsValue(q.Method(), execTime, q.Params())
 
 	if r.Error == nil {
-		c.service.logger.LogSuccessfulQuery(q.Method(), execTime, q.Params())
+		c.service.logger.LogSuccessfulQuery(q.Method(), execTime, q.paramsForLog())
```
tiger5226: A better convention for go is to check if `r.Error != nil` and remove the else altogether.

app/proxy/service_test.go
```diff
@@ -116,7 +116,29 @@ func TestCallResolve(t *testing.T) {
...
+
+	svc := NewService(config.GetLbrynet())
+	c := svc.NewCaller()
+	c.SetAccountID(acc.ID)
+
+	request := newRawRequest(t, "account_balance", nil)
+	hook := logrus_test.NewLocal(svc.logger.Logger())
+	c.Call(request)
+
+	assert.Equal(t, map[string]interface{}{"account_id": "****"}, hook.LastEntry().Data["params"])
```
tiger5226: what is the meaning of `***`? Is that a default account of sorts? I would probably put this into a constant so the code reads better.

internal/monitor/monitor.go
```diff
@@ -193,8 +208,8 @@ func (l *ProxyLogger) LogSuccessfulQuery(method string, time float64, params int
 func (l *ProxyLogger) LogFailedQuery(method string, params interface{}, errorResponse interface{}) {
 	l.entry.WithFields(logrus.Fields{
 		"method":   method,
-		"params":   params,
 		"response": errorResponse,
+		"params":   params,
```
tiger5226: why the move? alphabetical order is better.

app/proxy/service_test.go
sayplastic: That's just a masked value so we see the field is there but the value is not revealed. It is defined as a constant in the code, just double-checking its value here in the test.

sayplastic:
> Why not use something that already exists?

I'm actually using prometheus and its libraries. This code is to decouple main logic from those libraries so they can be replaced with something entirely else in the future, like new relic or metabase.

app/proxy/service.go
sayplastic: it wouldn't be true in two cases:

1. params is `nil`
2. if `map[string]string{}{"account_id": id}` was attached to params later. which I was doing. but I've changed that and now the first clause is no longer needed. thanks for prompting me to think more about this piece of code, I think it's improved now.

app/proxy/service.go
sayplastic: there is a non-returning pathway for when there is an error as well, so if/else is required. but yeah, replaced this with more familiar `!= nil`

PR #59, opened 2019-08-28
Feature/publish

sayplastic:


api/handlers.go
```diff
@@ -17,7 +17,7 @@ var Collector = metrics.NewCollector()
 
 // Index serves a blank home page
 func Index(w http.ResponseWriter, req *http.Request) {
-	w.WriteHeader(http.StatusOK)
+	http.Redirect(w, req, "https://lbry.tv/", http.StatusSeeOther)
```
nikooo777: by hardcoding the domain here we make it impossible for the project to be served elsewhere. perhaps if we really need to pass a URL we should have it in a config file

app/users/authenticator.go
```diff
@@ -0,0 +1,62 @@
...
+// Wrap result can be supplied to all functions that accept http.HandleFunc,
+// supplied function will be wrapped and called with AuthenticatedRequest instead of http.Request.
+func (a *Authenticator) Wrap(wrapped AuthenticatedFunc) http.HandlerFunc {
+	return func(w http.ResponseWriter, r *http.Request) {
+		AccountID, err := a.GetAccountID(r)
+		ar := &AuthenticatedRequest{Request: r}
+		if err != nil {
+			ar.AuthError = err
+		} else {
+			ar.AccountID = AccountID
```
nikooo777: can the accountID ever be nil or invalid? perhaps you need to check that too?

sayplastic: nah, it's either retrieved from the DB or it isn't, in that case the `err` will be set.

api/handlers.go
sayplastic: good point, I'll make it a setting

PR #66, opened 2019-09-03
Feature/publish

sayplastic:
This adds API endpoint for publishing media on lbrytv.

I tried to separate the endpoint logic into two entities:

- `UploadHandler.Handle`, which handles HTTP stuff, saves HTTP file and forwards UI JSON-RPC query to...
- `LbrynetPublisher`, which patches client request with necessary fields and forwards it to the running SDK instance and is currently mocked in endpoint tests

This is still work in progress but endpoint design is pretty much finalized.


app/publish/publish.go
```diff
@@ -0,0 +1,132 @@
...
+		}
+		w.WriteHeader(http.StatusOK)
+		w.Write(authErr.AsBytes())
+		return
+	}
+	file, header, err := r.FormFile("file")
+	if err != nil {
+		panic(err)
+	}
+	defer file.Close()
```
tiger5226: this is also ignoring an error. There is also a good article about why not to do this. https://www.joeshaw.org/dont-defer-close-on-writable-files/

I handle this in our go apps like this https://github.com/lbryio/internal-apis/blob/master/app/db/init.go#L167-L183

Its just a couple examples from one file but calling the wrapper with defer is better so the error is not ignored.

app/publish/publish.go
```diff
@@ -0,0 +1,132 @@
...
+	if err != nil {
+		panic(err)
+	}
+	return ioutil.TempFile(path, fmt.Sprintf("*_%v", origFilename))
+}
+
+func (h uploadHandler) preparePath(accountID string) (string, error) {
+	path := path.Join(h.uploadPath, accountID)
+	err := os.MkdirAll(path, os.ModePerm)
+	return path, err
```
tiger5226: `errors.Err(err)`

app/publish/publish_test.go
```diff
@@ -0,0 +1,125 @@
...
+
+	http.HandlerFunc(authenticator.Wrap(pubHandler.Handle)).ServeHTTP(rr, req)
+	response := rr.Result()
+
+	assert.Equal(t, http.StatusOK, response.StatusCode)
+	err = json.Unmarshal(rr.Body.Bytes(), &rpcResponse)
+	require.Nil(t, err)
+	assert.Equal(t, "authentication required", rpcResponse.Error.Message)
+	require.False(t, publisher.called)
+	// logEntry := hook.LastEntry()
```
tiger5226: I see generally a bunch of commented out code. Most likely from testing but I would remove these in the final product.

app/users/testing.go
```diff
@@ -0,0 +1,23 @@
...
+)
+
+// TestUserRetriever is a helper allowing to test API endpoints that require authentication
+// without actually creating DB records.
+type TestUserRetriever struct {
+	AccountID string
+	Token     string
+}
+
+// Retrieve returns AccountID set during TestUserRetriever creation,
```
tiger5226: You have the function comments here both a lot of other places there are no comments on exported functions, constants and vars.

api/routes.go
```diff
@@ -16,4 +19,10 @@ func InstallRoutes(ps *proxy.Service, r *mux.Router) {
 	r.HandleFunc("/api/proxy/", captureErrors(proxyServer.Handle))
 	r.HandleFunc("/content/claims/{uri}/{claim}/{filename}", captureErrors(ContentByClaimsURI))
 	r.HandleFunc("/content/url", captureErrors(ContentByURL))
+
+	actionsRouter := r.Path("/api/v1/actions").Subrouter()
```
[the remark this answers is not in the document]
sayplastic: Yeah, this will be improved. I'm not a fan of thick endpoint functions though so trying to find some readable way to keep dependencies/state outside.

app/publish/publish.go
```diff
@@ -0,0 +1,132 @@
...
+		FilePath                      string  `json:"file_path,omitempty"`
+		FileSize                      *string `json:"file_size,omitempty"`
+		IncludeProtoBuf               bool    `json:"include_protobuf"`
+		Blocking                      bool    `json:"blocking"`
+		*ljsonrpc.StreamCreateOptions `json:",flatten"`
+	}{}
+
+	query, err := proxy.NewQuery(rawQuery)
+	if err != nil {
+		panic(err)
```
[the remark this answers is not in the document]
sayplastic: Agreed, but this is still work in progress, hence panics in places they shouldn't be.

app/publish/publish.go
```diff
@@ -0,0 +1,132 @@
...
+
+	result, err := lbrynet.Client.StreamCreate(rpcParams.Name, filePath, bid, *rpcParams.StreamCreateOptions)
+	if err != nil {
+		return nil, err
+	}
+
+	rpcResponse := jsonrpc.RPCResponse{Result: result}
+	serialized, err := json.MarshalIndent(rpcResponse, "", "  ")
+	if err != nil {
+		return nil, err
```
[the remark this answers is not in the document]
sayplastic: If you mean `lbry.go`'s errors package, then that's a no go for me unfortunately, still cannot use it because errors appear as [robot vomit](https://github.com/lbryio/lbry.go/issues/62) in tests. Haven't gotten around to fixing it yet.

app/publish/publish.go
```diff
@@ -0,0 +1,132 @@
...
+func (h uploadHandler) Handle(w http.ResponseWriter, r *users.AuthenticatedRequest) {
+	if !r.IsAuthenticated() {
+		var authErr Error
+		if r.AuthFailed() {
+			authErr = NewAuthError(r.AuthError)
+		} else {
+			authErr = ErrUnauthorized
+		}
+		w.WriteHeader(http.StatusOK)
+		w.Write(authErr.AsBytes())
```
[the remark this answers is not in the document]
sayplastic: If an error occurs while we are writing to the HTTP response, there's not much we can do to recover or report anyway.

app/publish/publish.go
sayplastic: That's a very good point, thanks for bringing this to my attention

internal/lbrynet/publish.go
```diff
@@ -0,0 +1,80 @@
+package lbrynet
+
+type PublishParams struct {
```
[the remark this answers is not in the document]
sayplastic: Was eventually removed. We'll be using `stream_create` instead of `publish`

sayplastic:
This is no longer a WIP and ready to be deployed, I believe. Also contains some crucial auth token fixes.

app/publish/publish.go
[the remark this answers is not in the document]
tiger5226: I disagree. We should certainly log the error. At the very least niko or I can see the error log on the machine and will see what the error is. I don't think we should ever ignore errors.

app/publish/publish.go
```diff
@@ -1,131 +1,127 @@
...
-	if err != nil {
-		panic(err)
-	}
-	defer file.Close()
 
-	f, err := h.CreateFile(r.AccountID, header.Filename)
+	f, err := h.handleUpload(r)
 	if err != nil {
-		panic(err)
+		w.Write(NewInternalError(err).AsBytes())
```
tiger5226: This is an example of where I feel like we are not using a pattern that we have already defined with internal-apis. I see where you are adding code, now in multiple PRs that I have reviewed. I feel like the internal-apis pattern is a much better design and makes adding apis super easy. This is a lot of manual code handling. take a look at the api server package in lbry.go. I know you did not start from this ( not sure why grin didn't have you start from there ), but maybe you can leverage some of the code there to make your work easier. I see you writing to the responsewriters all the time. In the apiserver package we have a generic function that handles this for us so we never have to write a line like you do above. We have a special type for response and we can leverage the status automatically like `api.Response{Error:err, Status: http.StatusBadRequest}` and we can use things like `api.StatusError{Err:err,Status: http.StatusNotAuthenticated}` which returns an error and the code knows how to handle that. All regular errors are automatically server errors, and the stack traces get sent along with he response automatically etc. It just seems like a much better pattern. Not sure if you are aware of it. I think you would like it alot better.

app/users/users.go
```diff
@@ -32,21 +30,23 @@ type Retriever interface {
...
+	// s.updateLogger(monitor.F{})
 	return s
 }
 
-func (s *UserService) updateLogger(fields monitor.F) {
-	fields[monitor.TokenF] = s.token
-	s.log = monitor.NewModuleLogger("users").LogF(fields)
-}
+// func (s *UserService) updateLogger(fields monitor.F) {
+// 	fields[monitor.TokenF] = s.token
```
tiger5226: removed commented code

app/users/users.go
```diff
@@ -32,21 +30,23 @@ type Retriever interface {
...
+// }
 
 func (s *UserService) getLocalUser(id int) (*models.User, error) {
 	return models.Users(models.UserWhere.ID.EQ(id)).OneG()
 }
 
 func (s *UserService) createLocalUser(id int) (*models.User, error) {
+	log := s.logger.LogF(monitor.F{"id": id})
+
 	u := new(models.User)
```
tiger5226: I know this is not part of the PR but how come you use new instead of the sqlboiler referenced way `u := &models.User{}`?

app/publish/publish.go
```diff
@@ -1,131 +1,127 @@
...
+}
 
-	if rpcParams.FilePath != "__POST_FILE__" {
-		panic("unknown file_path content")
+func (h UploadHandler) handleUpload(r *users.AuthenticatedRequest) (*os.File, error) {
+	file, header, err := r.FormFile("file")
+	if err != nil {
+		return nil, err
 	}
+	defer file.Close()
```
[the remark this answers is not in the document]
sayplastic: The file is open for reading, why not

app/users/users.go
sayplastic: Aren't those two constructs essentially identical?

app/publish/publish.go
sayplastic: Grin put me on that path with his initial commit so I'm aware of that pattern. Unfortunately I had to abandon it straight away because one important distinction of my API endpoints is that they seldom generate the responses they're sending to the client. It's mostly errors, and they're not terribly complicated. Even in this package I'm simply passing down the wire the `[]byte` response I get from the SDK without any attempt to parse it. Hence I cannot have my views simply return `api.Response` structs.

I may be missing something but it seems for me to use `lbry.go/server` machinery here would be more trouble than it's worth.

app/publish/publish.go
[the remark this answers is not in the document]
sayplastic: Sorry, didn't see this comment behind all the lbry-bot emails. In my experience, 100% of errors writing to http socket are originating from naturally disconnecting clients and there's no reason to pay attention to that, there's just too much noise. If we experience some sort of persistent network failure, we're bound to learn about it in other ways.

PR #71, opened 2019-09-10
Feature/funding account

sayplastic:


sayplastic:
> LGTM, though you should run `go mod tidy`, I tried it locally and there's stuff to clean up.
> perhaps adding a check in CI could help detect these too (we do that in several projects)

I did that just three hours ago, the changes are a part of this PR. Strange

PR #75, opened 2019-09-16
Feature/unified endpoint

sayplastic:
The purpose of this is to avoid extra API endpoint for publishing and accept different kinds of params at a single URL:

- JSONRPC payload as HTTP body — for regular SDK requests

- JSONRPC payload as a multipart POST field + binary file as a multipart POST field — for `publish` requests


app/publish/publish.go
```diff
@@ -62,9 +86,9 @@ func (p *LbrynetPublisher) Publish(filePath, accountID string, rawQuery []byte)
 	return r
 }
 
-func (h UploadHandler) handleUpload(r *users.AuthenticatedRequest) (*os.File, error) {
+func (h UploadHandler) processUploadedFile(r *users.AuthenticatedRequest) (*os.File, error) {
```
lyoshenka: names like `ProcessSomething` or `HandleSomething` are not descriptive. what does it mean to process an uploaded file? what's actually happening? I would recommend something like `SaveFile` or `WriteFile` (Uploaded is redundant because it's already bound to an UploadHandler)

app/publish/publish.go
```diff
@@ -36,14 +39,35 @@ type LbrynetPublisher struct {
...
 // NewUploadHandler returns a HTTP upload handler object.
-func NewUploadHandler(uploadPath string, publisher Publisher) UploadHandler {
+func NewUploadHandler(opts UploadOpts) UploadHandler {
+	var (
+		publisher  Publisher
+		uploadPath string
+	)
+	if opts.ProxyService != nil {
+		publisher = &LbrynetPublisher{Service: opts.ProxyService}
+	} else if opts.Publisher != nil {
```
tiger5226: what if `ProxyService` and `opts.Publisher` are nil? then you are return an empty publisher with handler. Does this matter?

app/publish/publish.go
```diff
@@ -127,7 +150,18 @@ func (h UploadHandler) CreateFile(accountID string, origFilename string) (*os.Fi
...
 func (h UploadHandler) preparePath(accountID string) (string, error) {
-	path := path.Join(h.uploadPath, accountID)
+	path := path.Join(h.UploadPath, accountID)
 	err := os.MkdirAll(path, os.ModePerm)
 	return path, err
 }
+
+// CanHandle checks if http.Request contains POSTed data in an accepted format.
+// Supposed to be used in gorilla mux router MatcherFunc.
+func (h UploadHandler) CanHandle(r *http.Request, _ *mux.RouteMatch) bool {
```
tiger5226: A better name would be `HasAcceptableFormat`

app/publish/publish.go
```diff
@@ -127,7 +150,18 @@ func (h UploadHandler) CreateFile(accountID string, origFilename string) (*os.Fi
...
+	_, _, err := r.FormFile(FileFieldName)
```
tiger5226: What happens if it's a different error? What then? We assume we can handle it?

app/publish/testing.go
```diff
@@ -0,0 +1,38 @@
...
+
+// CreatePublishRequest creates and returns a HTTP request providing data for the publishing endpoint.
+func CreatePublishRequest(t *testing.T, data []byte) *http.Request {
+	readSeeker := bytes.NewReader(data)
+	body := &bytes.Buffer{}
+
+	writer := multipart.NewWriter(body)
+
+	fileBody, err := writer.CreateFormFile(FileFieldName, "lbry_auto_test_file")
+	require.Nil(t, err)
```
tiger5226: I like this! I need to start using this instead.

app/publish/publish.go
```diff
@@ -93,7 +117,6 @@ func (h UploadHandler) handleUpload(r *users.AuthenticatedRequest) (*os.File, er
 // in a mux.Router.
 func (h UploadHandler) Handle(w http.ResponseWriter, r *users.AuthenticatedRequest) {
 	w.WriteHeader(http.StatusOK)
```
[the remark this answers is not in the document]
sayplastic: JSONRPC doesn't make use of HTTP status codes, it's always 200.

app/publish/publish.go
sayplastic: If the field is present, we assume we _should_ handle it. It is the job of the main handler routine to deal with whatever other errors might occur.

app/publish/publish.go
sayplastic: In this case it will be initialized with an empty publisher and that will bite in the rear when someone tries to publish a file with this instance. It's not a good idea to set a default there.

I thought I'd rather let it fail in case of a programmer's error like this but then I found examples of constructors returning  `*Val, error` in the stdlib so I'll fix it to do that.

app/publish/publish.go
sayplastic: There is a point but I think I'd rather keep the name describing _the what_ (we're checking if we can handle the request) instead of disclosing _the how_ (we're checking the format of the supplied request because..?). Hope it makes sense :)

sayplastic:
> I see a few `panic()`s in your code. You should basically never use panic. If there's an error, either handle it or pass it up the call chain to be handled elsewhere.

There was a leftover from debugging in the publish module, thanks for bringing it to my attention. I would normally only use `panic` in tests or in cases of unrecoverable errors, such as db or config initialization problems.

app/publish/testing.go
sayplastic: Yeah, it's quite handy. One problem is it doesn't handle printing `lbry.go` custom errors very well (I mentioned that before). I'll get around to fixing it one day.

PR #76, opened 2019-09-17
Feature/reflection

sayplastic:
Changed to use public API. I decided that to make any optimizations like parallel uploads at this point would be premature.


deployments/docker/app/config/lbrytv.yml
```diff
@@ -15,3 +15,6 @@ MetricsAddress: :2112
 MetricsPath: /metrics
 
 PublishSourceDir: /storage/publish
+BlobFilesDir: /storage/lbrynet/blobfiles
+
+Reflector: reflector.lbry.com:5566
```
lyoshenka: I would call this ReflectorURL

internal/reflection/manager.go
```diff
@@ -0,0 +1,163 @@
...
+				}
+			}
+		}
+	}()
+}
+
+// Abort resets the upload schedule and cancels blob upload
+// after the currently uploading blob is finished.
+func (r *Manager) Abort() {
+	r.abortTimer <- true
```
lyoshenka: have you looked at https://github.com/lbryio/lbry.go/tree/master/extras/stop ?

deployments/docker/app/config/lbrytv.yml
sayplastic: That might be misleading as it's technically not a URL

lyoshenka: You can call it `ReflectorHostPort` if you want to be very precise

sayplastic: Renamed to `ReflectorAddress`.

internal/reflection/manager.go
sayplastic: I have, but cannot figure out how it would be beneficial to use here instead of two plain channels. Might be a different story if uploads need to be parallelized later.

PR #79, opened 2019-09-19
Feature/real ip

sayplastic:


app/users/authenticator.go
```diff
@@ -29,7 +29,7 @@ func NewAuthenticator(retriever Retriever) *Authenticator {
 // an SDK account ID from Retriever.
 func (a *Authenticator) GetAccountID(r *http.Request) (string, error) {
 	if token, ok := r.Header[TokenHeader]; ok {
```
nikooo777: i'm not 100% sure but I think you shouldn't be accessing the headers this way. instead you should use 
r.GetHeader(TokenHeader) which returns a string.

app/users/helpers.go
```diff
@@ -0,0 +1,94 @@
+package users
+
+import (
+	"bytes"
+	"net"
+	"net/http"
+	"strings"
+)
+
+// most of this is from https://husobee.github.io/golang/ip-address/2015/12/17/remote-ip-go.html
```
nikooo777: i've seen this elsewhere, I think internal-apis.
perhaps it'd be time to unify this in lbry.go?

app/users/users.go
```diff
@@ -30,7 +30,13 @@ type savedAccFields struct {
...
 // Retriever is an interface for user retrieval by internal-apis auth token
 type Retriever interface {
-	Retrieve(token string) (*models.User, error)
+	Retrieve(query Query) (*models.User, error)
+}
+
+// Query contains queried user details and optional metadata about the request
+type Query struct {
+	Token        string
+	MetaRemoteIP string
```
nikooo777: why "meta"?

app/users/users.go
```diff
@@ -72,11 +78,12 @@ func (s *UserService) createDBUser(id int) (*models.User, error) {
 }
 
 // Retrieve authenticates user with internal-api and retrieves/creates locally stored user.
-func (s *UserService) Retrieve(token string) (*models.User, error) {
+func (s *UserService) Retrieve(q Query) (*models.User, error) {
+	token := q.Token
 	log := s.logger.LogF(monitor.F{"token": token})
```
nikooo777: aren't we leaking tokens this way? is this truly necessary?

app/users/users.go
```diff
@@ -147,9 +154,9 @@ func (s *UserService) saveAccFields(accFields savedAccFields, u *models.User) er
 
 // GetAccountIDFromRequest retrieves SDK  account_id of a user making a http request
 // by a header provided by http client.
-func GetAccountIDFromRequest(req *http.Request, retriever Retriever) (string, error) {
-	if token, ok := req.Header[TokenHeader]; ok {
-		u, err := retriever.Retrieve(token[0])
+func GetAccountIDFromRequest(r *http.Request, retriever Retriever) (string, error) {
+	if token, ok := r.Header[TokenHeader]; ok {
```
nikooo777: same as above (r.GetHeader())

app/users/users.go
sayplastic: It's metadata for the query, does not affect the query itself.

app/users/users.go
sayplastic: We are not, tokens are filtered out in the logging package in production. It's necessary for debugging / tests.

app/users/authenticator.go
sayplastic: There's no GetHeader method on `http.Request` and nothing mentioned in the docs.

app/users/helpers.go
sayplastic: Yeah, I copied it from internal-apis, which I cannot reuse in public code. One more strike and I'm moving it out.

PR #88, opened 2019-10-03
Feature/wallets migrate

sayplastic:
This is an accounts -> wallets migration script, built upon the PR #85.

### DIFF
--- a/cmd/migrate_to_wallets.go
+++ b/cmd/migrate_to_wallets.go
@@ -34,8 +34,7 @@ var migrateToWallets = &cobra.Command{
 		}
 		c := lbrynet.Client
 
-		usersQ := models.Users(models.UserWhere.WalletID.EQ(""))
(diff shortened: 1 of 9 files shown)


cmd/migrate_to_wallets.go
```diff
@@ -0,0 +1,112 @@
...
+	Hidden: true,
+	Run: func(cmd *cobra.Command, args []string) {
+		realRun := false
+		if len(args) > 0 && args[0] == "doit" {
+			realRun = true
+		}
+		c := lbrynet.Client
+
+		usersQ := models.Users(models.UserWhere.WalletID.EQ(""))
+		users, err := usersQ.AllG()
```
lyoshenka: i usually do this as one line: `users, err := models.Users(models.UserWhere.WalletID.EQ("")).AllG()`

PR #99, opened 2019-10-21
Bugfix/wallet

sayplastic:


sayplastic:
Added some crucial changes, although not purely wallet-related. Massively improved how metrics are collected throughout the app.

app/users/wallets.go
```diff
@@ -55,33 +56,52 @@ func (s *WalletService) Retrieve(q Query) (*models.User, error) {
...
 			return nil, err
 		}
+		doWalletInit = false
 		err := s.saveWalletID(localUser, wid)
 		if err != nil {
 			return nil, err
 		}
 	}
 
+	if doWalletInit {
```
sayplastic: Completely forgotten about wallet handling improvements here, worth a review as well.

app/proxy/service.go
```diff
@@ -251,17 +249,17 @@ func (c *Caller) call(rawQuery []byte) (*jsonrpc.RPCResponse, CallError) {
 
 	queryStartTime := time.Now()
 	r, err := c.sendQuery(q)
+	duration := time.Now().Sub(queryStartTime).Seconds()
```
nikooo777: why not time.since(queryStartTime) ?

app/users/remote.go
```diff
@@ -14,14 +18,21 @@ type RemoteUser struct {
...
 	u := &RemoteUser{}
 	c := lbryinc.NewClient(token, &lbryinc.ClientOpts{
-		ServerAddress: config.GetInternalAPIHost(), RemoteIP: remoteIP})
+		ServerAddress: config.GetInternalAPIHost(),
+		RemoteIP:      remoteIP,
+	})
 
+	start := time.Now()
 	r, err := c.UserMe()
+	duration := time.Now().Sub(start).Seconds()
```
nikooo777: same as above

app/proxy/service.go
sayplastic: forgot about this shortcut

sayplastic:
> Could a user accidentally get 2 wallets initialized for his person if he hits the API twice at the same time? 

Given that wallets are stored as files named after user IDs, I don't see a way this could happen.

PR #107, opened 2019-10-30
Fix docker-compose command in readme

g1tman:
`docker-compose up --no-start 'services'`
instead of
`docker-compose up 'services' --no-start`

The later one will give an error
`ERROR: No such service: --no-start`


tzarebczan:
Thanks for the PR @g1tman ! Can we show you some [appreciation](https://lbry.com/faq/appreciation)?

PR #110, opened 2019-11-01
Fix endless loop video playback error

sayplastic:
This fixes an off-by-one video playback error which caused endless blob download loop for small streams.

### DIFF
--- a/app/player/player.go
+++ b/app/player/player.go
@@ -257,9 +257,6 @@ func (s *reflectedStream) streamBlob(blobNum int, startOffsetInBlob int64, dest 
 	logBlobNum := fmt.Sprintf("%v/%v", bi.BlobNum+1, s.blobNum())
 
 	readLen := 0
-	// if n > 0 {
-	// 	startOffsetInBlob = 0
-	// }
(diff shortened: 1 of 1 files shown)


app/player/player.go
```diff
@@ -238,13 +252,14 @@ func (s *reflectedStream) getBlob(url string) (*http.Response, error) {
...
+func (s *reflectedStream) streamBlob(blobNum int, startOffsetInBlob int64, dest []byte) (int, error) {
 	bi := s.SDBlob.BlobInfos[blobNum]
-	logBlobNum := fmt.Sprintf("%v/%v", bi.BlobNum, len(s.SDBlob.BlobInfos))
+	logBlobNum := fmt.Sprintf("%v/%v", bi.BlobNum+1, s.blobNum())
 
-	if n > 0 {
-		startOffsetInBlob = 0
-	}
+	readLen := 0
+	// if n > 0 {
```
nikooo777: cleanup needed?

sayplastic: Yep, sorry about that!

PR #114, opened 2019-11-13
Add ability for multiple lbrynet server connections and routing.

tiger5226:


app/router/router_test.go
```diff
@@ -0,0 +1,43 @@
...
+	server = sdkRouter.GetSDKServer("lbrytv-id.751365.wallet")
+	assert.Equal(t, "http://lbrynet3:5279/", server)
+}
+
+func TestServerRetrieval(t *testing.T) {
+	sdkRouter := New(config.GetAllLbrynets())
+	servers := sdkRouter.GetSDKServerList()
+	for i := 0; i < 10000; i++ {
+		iStr := strconv.Itoa(i)
+		server := sdkRouter.GetSDKServer("lbrytv-id." + iStr + ".wallet")
```
sayplastic: Can use `lbrynet.MakeWalletID` for creating properly formatted wallet IDs.

app/router/router.go
```diff
@@ -0,0 +1,116 @@
+package router
```
sayplastic: Perhaps this can live in the lbrynet package — I was going to delegate all lbrynet server-related communications to that module eventually anyway.

app/router/router.go
```diff
@@ -0,0 +1,116 @@
...
+		LbrynetServers: sdkServers,
+		logger:         monitor.NewModuleLogger("router"),
+	}
+}
+
+func NewDefault() SDKRouter {
+	return SDKRouter{LbrynetServers: map[string]string{
+		"default": config.GetLbrynet(),
+	},
+		logger: monitor.NewModuleLogger("router"),
```
sayplastic: Some funky formatting here

app/router/router.go
```diff
@@ -0,0 +1,116 @@
...
+}
+
+func (r *SDKRouter) GetSDKServer(walletID string) string {
+	sdk := r.balancedSDK()
+	if walletID != "" {
+		sdk = r.getSDKByWalletID(walletID)
+	}
+
+	if walletID != "" && sdk == "" {
+		r.logger.Log().Info("walletID is set but there is no server associated with it.")
```
sayplastic: This should be an error

app/router/router.go
```diff
@@ -0,0 +1,116 @@
...
+		sdk = r.defaultSDKServer() //"UNKOWN_SERVER"
+	}
+	r.logger.Log().Infof("From wallet id [%s] server [%s] is returned", walletID, sdk)
```
sayplastic: `Debugf`

app/router/router.go
```diff
@@ -0,0 +1,116 @@
...
+}
+
+func (r *SDKRouter) getSDKByWalletID(walletID string) string {
+	if len(r.servers) == 0 {
+		r.sortServers()
+	}
+	digit := getLastDigit(walletID)
+	server := int64(0)
+	if len(r.servers) == 0 {
+		r.logger.Log().Info("There are no servers listed. Something went terribly wrong.")
```
sayplastic: That's an `Error` as well

app/router/router.go
```diff
@@ -0,0 +1,116 @@
...
+
+func New(sdkServers map[string]string) SDKRouter {
+	return SDKRouter{
+		LbrynetServers: sdkServers,
+		logger:         monitor.NewModuleLogger("router"),
+	}
+}
+
+func NewDefault() SDKRouter {
+	return SDKRouter{LbrynetServers: map[string]string{
```
sayplastic: I feel like this should call `New` just for consistency.

app/router/router.go
```diff
@@ -0,0 +1,116 @@
...
+	if walletID != "" && sdk == "" {
+		r.logger.Log().Info("walletID is set but there is no server associated with it.")
+		sdk = r.defaultSDKServer() //"UNKOWN_SERVER"
+	}
+	r.logger.Log().Infof("From wallet id [%s] server [%s] is returned", walletID, sdk)
+	return sdk
+}
+
+func (r *SDKRouter) getSDKByWalletID(walletID string) string {
+	if len(r.servers) == 0 {
```
sayplastic: Why don't we sort them servers once in the `New()` and only check here the list length for sanity?

app/router/router.go
```diff
@@ -0,0 +1,116 @@
...
+type SDK struct { //Remove this once we can store names in DB
+	Name    string
+	Address string
+}
+
+func SingleSDKSet(address string) map[string]string {
+	return map[string]string{"default": address}
+}
+
+func New(sdkServers map[string]string) SDKRouter {
```
sayplastic: I would cover this `New` function with a unit test. If we move server sorting stuff here, it will make sense to cover in the same test as well.

PR #116, opened 2019-11-20
Improvement/player

sayplastic:
Streaming directly from reflector service implemented.
Most improvements are in `app/player/player.go` which has been completely rewritten.


app/player/player.go
```diff
@@ -21,17 +21,18 @@ import (
 	"github.com/lbryio/lbry.go/v2/stream"
 )
 
-const reflectorURL = "http://blobs.lbry.io/"
+const reflectorURL = "http://blobs.lbry.io"
```
nikooo777: just so you're aware, this isn't reflector, this is the actual blob storage.
reflector would be reflector.lbry.com

internal/monitor/monitor.go
```diff
@@ -135,7 +135,7 @@ func LogFailedQuery(method string, query interface{}, errorResponse interface{})
 		"response": errorResponse,
 	}).Error("daemon responded with an error")
 
-	captureFailedQuery(method, query, errorResponse)
+	// captureFailedQuery(method, query, errorResponse)
```
nikooo777: might want to remove it if you're not using it?

internal/monitor/monitor.go
```diff
@@ -195,7 +195,7 @@ func (l *ProxyLogger) LogFailedQuery(method string, params interface{}, errorRes
 		"response": errorResponse,
 	}).Error("error from the target endpoint")
 
-	captureFailedQuery(method, params, errorResponse)
+	// captureFailedQuery(method, params, errorResponse)
```
nikooo777: might want to remove it if you're not using it 2

internal/reflection/manager.go
```diff
@@ -9,7 +9,7 @@ import (
 	"github.com/lbryio/lbrytv/internal/monitor"
 
 	"github.com/lbryio/lbry.go/extras/errors"
-	"github.com/lbryio/lbry.go/stream"
+	stream "github.com/lbryio/lbry.go/stream"
```
tiger5226: is this really needed?

server/server.go
```diff
@@ -22,31 +22,38 @@ var logger = monitor.NewModuleLogger("server")
 type Server struct {
 	monitor.ModuleLogger
 
-	InterruptChan  chan os.Signal
 	DefaultHeaders map[string]string
 	Environment    *environment.Env
 	ProxyService   *proxy.ProxyService
 
+	stopChan chan os.Signal
```
tiger5226: We have a stopper pattern in lbry.go https://github.com/lbryio/lbry.go/tree/master/extras/stop

You may find it useful. I use it in both internal-apis and chainquery. It's pretty useful.

go.mod
```diff
@@ -3,7 +3,7 @@ module github.com/lbryio/lbrytv
 require (
 	github.com/aws/aws-sdk-go v1.23.19 // indirect
 	github.com/btcsuite/btcd v0.0.0-20190824003749-130ea5bddde3 // indirect
-	github.com/getsentry/sentry-go v0.3.0
+	github.com/getsentry/sentry-go v0.3.1
```
tiger5226: just making sure this was intended

server/server.go
sayplastic: I looked at it but felt like it's an overkill here. Plus we don't have direct control over whatever goroutines are spawned during the server life cycle.

go.mod
sayplastic: Yep, I'm keeping this dependency up to date whenever possible.

sayplastic:
This needs further improvements.

PR #123, opened 2019-12-05
Do not authenticate method calls that do not require account identifi…

sayplastic:
Lbrytv was authenticating every call whenever auth token header was present, even for methods that did not require it. This fixes it.

### DIFF
--- a/app/proxy/handlers.go
+++ b/app/proxy/handlers.go
@@ -40,18 +40,21 @@ func (rh *RequestHandler) Handle(w http.ResponseWriter, r *http.Request) {
 
 	var walletID string
 	if config.AccountsEnabled() {
-		retriever := users.NewWalletService()
-		auth := users.NewAuthenticator(retriever)
-		walletID, err = auth.GetWalletID(r)
+		q, err := NewQuery(body)
+		if err != nil || !methodInList(q.Method(), relaxedMethods) {
+			retriever := users.NewWalletService()
+			auth := users.NewAuthenticator(retriever)
+			walletID, err = auth.GetWalletID(r)
 
-		// TODO: Refactor error response creation
-		if err != nil {
-			response, _ := json.Marshal(NewErrorResponse(err.Error(), ErrAuthFailed))
-			w.Header().Set("Content-Type", "application/json; charset=utf-8")
-			w.WriteHeader(http.StatusOK)
-			w.Write(response)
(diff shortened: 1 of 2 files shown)


app/proxy/handlers.go
```diff
@@ -40,18 +40,21 @@ func (rh *RequestHandler) Handle(w http.ResponseWriter, r *http.Request) {
 
 	var walletID string
 	if config.AccountsEnabled() {
-		retriever := users.NewWalletService()
-		auth := users.NewAuthenticator(retriever)
-		walletID, err = auth.GetWalletID(r)
+		q, err := NewQuery(body)
```
tiger5226: I thought all accounts logic is no longer in use?

tiger5226: I thought all accounts logic is no longer in use?

app/proxy/handlers.go
```diff
@@ -40,18 +40,21 @@ func (rh *RequestHandler) Handle(w http.ResponseWriter, r *http.Request) {
...
-			response, _ := json.Marshal(NewErrorResponse(err.Error(), ErrAuthFailed))
-			w.Header().Set("Content-Type", "application/json; charset=utf-8")
-			w.WriteHeader(http.StatusOK)
-			w.Write(response)
-			monitor.CaptureRequestError(err, r, w)
-			return
+			// TODO: Refactor error response creation
+			if err != nil {
+				response, _ := json.Marshal(NewErrorResponse(err.Error(), ErrAuthFailed))
+				w.Header().Set("Content-Type", "application/json; charset=utf-8")
```
tiger5226: maybe its time to refactor this https://github.com/lbryio/lbrytv/search?q=application%2Fjson&unscoped_q=application%2Fjson

sayplastic: Agree, it's in the TODO as well.

app/proxy/handlers.go
sayplastic: What accounts logic?

tiger5226: I think  maybe its just a naming issue, but I was referring to the `if config.AccountsEnabled()`.

sayplastic: Right, that was a naming discrepancy. I just removed the config flag altogether since there's no reason to support both logical pathways anymore.

Issue #124, opened 2019-12-11
Create status endpoint for lbrytv


andybeletsky:
- Create status endpoint for lbrytv
- Report abnormal app shutdowns

lyoshenka:
fixed by #161

andybeletsky:
Done: https://api.lbry.tv/internal/status

PR #126, opened 2019-12-12
Improvement/player

sayplastic:
Massive improvements of the player package:

- Refactored to be able to extend the code with local blob storage/caching
- Changed blob retrieval method to native reflector calls instead of using S3 storage directly
- Changed the way binary data is read from blobs physically, now using default `ioutil.CopyN` 32K buffer instead of hacked version using 2M buffer
- Implemented local storage for downloaded/decrypted blobs (aka cache)


app/player/cache.go
```diff
@@ -0,0 +1,199 @@
...
+
+func (s fsStorage) open(hash interface{}) (*os.File, error) {
+	f, err := os.Open(s.getPath(hash))
+	if err != nil {
+		return nil, err
+	}
+	return f, nil
+}
+
+// Has returns true if cache contains the requested chink.
```
tiger5226: chunk

app/player/cache.go
```diff
@@ -0,0 +1,199 @@
...
+// Get returns ReadableChunk if it can be retrieved from the cache by the requested hash
+// and a boolean representing whether chunk was found or not.
+func (c *fsCache) Get(hash string) (ReadableChunk, bool) {
+	if value, ok := c.rCache.Get(hash); ok {
+		f, err := c.storage.open(value)
+		if err != nil {
+			metrics.PlayerCacheErrorCount.Inc()
+			CacheLogger.Log().Errorf("chunk %v found in cache but couldn't open the file: %v", hash, err)
+			c.rCache.Del(value)
+			return nil, false
```
tiger5226: how is it returning nil, if your not returning a pointer.

app/player/cache.go
```diff
@@ -0,0 +1,199 @@
...
+		}
+		metrics.PlayerCacheHitCount.Inc()
+		cb, err := initCachedChunk(f)
+		f.Close()
```
tiger5226: should you just `defer f.Close()`

app/player/cache.go
```diff
@@ -0,0 +1,199 @@
...
+		if err != nil {
+			CacheLogger.Log().Errorf("chunk %v found in cache but couldn't read the file: %v", hash, err)
```
tiger5226: how come we don't just roll up the error instead of returning true/false?

app/player/cache.go
```diff
@@ -0,0 +1,199 @@
...
+		numWritten, err := f.Write(body)
+		f.Close()
+		if err != nil {
+			metrics.PlayerCacheErrorCount.Inc()
+			CacheLogger.Log().Errorf("error saving cache file %v: %v", chunkPath, err)
+			return nil, err
+		}
+		CacheLogger.Log().Debugf("written %v bytes for chunk %v", numWritten, hash)
+	}
+	c.rCache.Set(hash, hash, int64(cacheCost))
```
tiger5226: if there was an error on `Set` we are still increasing the cache size. This gets things out of sync.

app/player/cache.go
```diff
@@ -0,0 +1,199 @@
...
+	// metrics.PlayerCacheSize.Set(float64(c.rCache.Metrics.CostAdded() - c.rCache.Metrics.CostEvicted()))
+
+	metrics.PlayerCacheSize.Add(float64(cacheCost))
+	CacheLogger.Log().Debugf("chunk %v successfully cached", hash)
```
tiger5226: we could also he sending a false message to the logger if the set fails.

app/player/cache.go
```diff
@@ -0,0 +1,199 @@
...
+	return nil, false
+}
+
+// Set takes chunk body and saves reference to it into the cache table
+func (c *fsCache) Set(hash string, body []byte) (ReadableChunk, error) {
+	cacheCost := len(body)
+
+	CacheLogger.Log().Debugf("attempting to cache chunk %v", hash)
+	chunkPath := c.storage.getPath(hash)
+	f, err := os.OpenFile(chunkPath, os.O_WRONLY|os.O_CREATE|os.O_EXCL, 0600)
```
tiger5226: your only handling one type of error with `os.IsExists`

app/player/cache.go
```diff
@@ -0,0 +1,199 @@
...
+	c.rCache.Set(hash, hash, int64(cacheCost))
+	// metrics.PlayerCacheSize.Set(float64(c.rCache.Metrics.CostAdded() - c.rCache.Metrics.CostEvicted()))
+
+	metrics.PlayerCacheSize.Add(float64(cacheCost))
+	CacheLogger.Log().Debugf("chunk %v successfully cached", hash)
+	return &cachedChunk{reflectedChunk{body}}, nil
+}
+
+// Remove deletes both cache record and chunk file from the filesystem.
+func (c *fsCache) Remove(hash string) {
```
tiger5226: This should return an error if it fails to remove. Also the things inside should return errors to check.

app/player/player.go
```diff
@@ -2,324 +2,478 @@ package player
...
+	return nil
 }
 
-func (s reflectedStream) blobNum() int {
-	return len(s.SDBlob.BlobInfos) - 1
+// ResolveStream resolves provided URI by calling the SDK.
+func (p *Player) ResolveStream(uri string) (*Stream, error) {
+	s := &Stream{URI: uri}
+
+	r, err := p.getLbrynetClient().Resolve(uri)
```
tiger5226: is a new player created for each handler? We should make sure we are getting a balanced SDK for each call otherwise it could get unbalanced if it continuously reuses the same sdk from the creation of the player object.

app/player/server.go
```diff
@@ -21,7 +19,7 @@ import (
 // If dst implements the ReaderFrom interface,
 // the copy is implemented using it.
 func CopyN(dst io.Writer, src io.Reader, n int64) (written int64, err error) {
-	buf := make([]byte, stream.MaxBlobSize)
+	buf := make([]byte, 1024*256)
```
tiger5226: magic number?

deployments/docker-production/docker-compose-scalable.yml
```diff
@@ -0,0 +1,171 @@
+version: '3.2'
```
tiger5226: we should remove this now since its in production.

deployments/docker-production/docker-compose.yml
```diff
@@ -21,6 +21,7 @@ services:
         awslogs-group: lbrytv
         awslogs-region: us-east-1
         awslogs-stream: app
+    restart: unless-stopped
```
tiger5226: do we have a mechanism to capture this? It will no longer be as obvious but it will be good to address these situations when it fails.

deployments/docker/app/config/lbrytv.yml
```diff
@@ -18,4 +18,9 @@ PublishSourceDir: /storage/publish
 BlobFilesDir: /storage/lbrynet/blobfiles
 
 ReflectorAddress: reflector.lbry.com:5566
+ReflectorTimeout: 10
```
tiger5226: maybe add a comment to show the unit of measure.

go.sum.orig
```diff
@@ -0,0 +1,669 @@
+cloud.google.com/go v0.26.0/go.mod h1:aQUYkXzVsufM+DwF1aE+0xfcU+56JwCaLick0ClmMTw=
```
tiger5226: not sure what this is doing here.

internal/monitor/module_logger.go
```diff
@@ -0,0 +1,71 @@
...
+		ModuleName: moduleName,
+		Logger:     logger,
+		Level:      logger.GetLevel(),
+	}
+	l.Logger.SetLevel(l.Level)
+	return l
+}
+
+// LogF is a deprecated method, an equivalent WithFields/WithField should be used.
+func (l ModuleLogger) LogF(fields F) *logrus.Entry { return l.WithFields(fields) }
```
tiger5226: why deprecate it? this isn't a library. Its an internal repo. We should just remove it no?

internal/reflection/manager.go
```diff
@@ -9,7 +9,7 @@ import (
 	"github.com/lbryio/lbrytv/internal/monitor"
 
 	"github.com/lbryio/lbry.go/extras/errors"
-	"github.com/lbryio/lbry.go/stream"
+	"github.com/lbryio/lbry.go/v2/stream"
```
tiger5226: we should remove usage of v1 brother.

app/player/cache.go
sayplastic: It's better to handle this as a cache miss and log it for later analysis instead of bubbling up the error which would eventually result in a stream interruption.

app/player/cache.go
sayplastic: Good point, although there's no error, it's just returning whether the item was admitted into cache or not.

app/player/player.go
sayplastic: `getLbrynetClient` just returns a balanced server on every call

app/player/server.go
sayplastic: This should really be removed

deployments/docker-production/docker-compose.yml
sayplastic: Only by visual log inspection currently. This should be fixed as a part of #124

app/player/cache.go
sayplastic: The function is returning an interface, which can be nil

app/player/handlers.go
```diff
@@ -0,0 +1,73 @@
...
+	return fmt.Sprintf("%s#%s", vars["uri"], vars["claim"])
+}
+
+func (h RequestHandler) writeErrorResponse(w http.ResponseWriter, s int, msg string) {
+	w.WriteHeader(s)
+	w.Write([]byte(msg))
+}
+
+func (h RequestHandler) processStreamError(w http.ResponseWriter, uri string, err error) {
+	if err == errPaidStream {
```
lyoshenka: if you're on go 1.13, start using `errors.Is` instead

https://blog.golang.org/go1.13-errors

sayplastic:
@lyoshenka,

> huge PR. I skimmed it because @tiger5226 already did a great job with the in-depth review.

This is a common complaint but I don't know what to do with it. Regardless of PRs size, by the time they are usually reviewed, my work on any particular piece of code has long moved forward and/or is deployed already. 

[Case in point](https://github.com/lbryio/lbrytv/pull/123/files), there's just 60 lines of code and it's been open for 9 days now. I'm not complaining, I know we are all very busy but that's a chronic pain point in our code reviews.

app/player/cache.go
sayplastic: Slightly changed the calls order so I can close it right away.

app/player/cache.go
tiger5226: I actually recently ran into this fact on lighthouse. I had the same question and realized it was an interface. I will look out of that in future PRs.

app/player/cache.go
tiger5226: That seems like a lesser quality pattern. I think the reasoning behind defer is to avoid bugs. Someone, or you might change code in the future and then they get a panic when its closed already because the generally accepted pattern of deferring wasn't followed. Something to consider.

tiger5226:
I am going to start to make it a priority to check lbry.tv PRs brother. I have never really looked for PRs to review here which is a big part of the problem. I only look when you ask me to. 

As to the size of the PRs, it's definitely a problem. I can assure you, that you will get a less thorough review from 99% coders as a result. 

As a general practice, I require my team to submit PRs for units of work that take 0.5, 1hr, 2hrs, 4hrs, or 8 hrs. No one is allowed to submit 16 hrs of work for review. In other words, if you worked a day without sending a PR in to be merged. One of the following things happened: 

- you forgot to send in the PR at the end of the day for work completed.
- you did not spend enough time breaking your work out into logical working units.
- you ignored the rules

In addition to this, the flip side is that a PR is never to be outstanding for more than 24 hrs, and most should be reviewed within 2-6 hrs depending on the time of day it is submitted. The first task in the morning for those on my team is to review PRs, then code. The thought process is that any code review left outstanding is a timeline cost. timeline for projects, timeline for feature releases, timeline for phases etc. 

While I think these rules are great for overall productivity, I am not trying to dictate, but at the very least, I should be dogfooding my beliefs and get your PRs reviewed. I will do a better job of that.

PR #133, opened 2019-12-20
Feature/downloadable

sayplastic:
This enables browser to display "Save as" dialog window for unsupported file types.

### DIFF
--- a/app/player/handlers.go
+++ b/app/player/handlers.go
@@ -7,8 +7,18 @@ import (
 	"strings"
 
 	"github.com/gorilla/mux"
+	"github.com/lbryio/lbrytv/app/users"
 	"github.com/lbryio/lbrytv/internal/monitor"
 )
+
+var viewableTypes = []string{
+	"audio/",
+	"video/",
+	"image/",
+	"text/markdown",
+}
+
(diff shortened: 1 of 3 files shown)


app/player/handlers.go
```diff
@@ -45,10 +55,41 @@ func (h RequestHandler) processStreamError(w http.ResponseWriter, uri string, er
...
+	err = h.player.RetrieveStream(s)
+	if err != nil {
+		Logger.streamRetrievalFailed(uri, err)
+		h.processStreamError(w, uri, err)
+		return
+	}
+	Logger.streamRetrieved(s)
+
+	if !h.isViewable(s.ContentType) || r.URL.Query().Get(ParamDownload) != "" {
+		w.Header().Set("Content-Disposition", fmt.Sprintf("attachment; filename=%v", s.Claim.Value.GetStream().Source.Name))
```
nikooo777: just wondering, could this open up to XSS or some sort of attack vector where people add arbitrary code as `name`?

sayplastic: This is not executed, it just determines the file name on the user's filesystem.

PR #138, opened 2019-12-31
Set wallet option skip_on_startup to true

tiger5226:
Per @lex this should be passed as true to prevent the wallet from being added to the config.


sayplastic:
Fixed CI config so the tests are passing now.

PR #140, opened 2020-01-02
Improvement/cache tweaks

sayplastic:
This mostly improves logging and adds a couple of metrics to better monitor cache behavior.

I also upped disk cache size to 32Gb.

### DIFF
--- a/app/player/player.go
+++ b/app/player/player.go
@@ -25,8 +25,12 @@ import (
 )
 
 const (
-	// ChunkSize is a size of decrypted blob
+	// ChunkSize is a size of decrypted blob.
 	ChunkSize = stream.MaxBlobSize - 1
+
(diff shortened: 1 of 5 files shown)


app/player/player.go
```diff
@@ -384,46 +402,51 @@ func (b *chunkGetter) prefetchToLocalCache(startN int) {
 	chunksLeft := len(b.sdBlob.BlobInfos) - startN - 1 // Last blob is empty
 	if chunksLeft <= 0 {
 		return
-	} else if chunksLeft > 10 {
-		prefetchLen = 10
+	} else if chunksLeft > 3 {
```
lyoshenka: this should prolly be a const, with some explanation of why the value is 3 or 10 or whatever

PR #144, opened 2020-01-04
DB Retry Behavior

tiger5226:
Don't just error on db connections. DBs lose connections, it happens. Retry periodically, with increasing intervals before finally returning an error to connect to the database.

This is also beneficial for our container compositions. If postgres does not startup in time, it will retry.

Screen Shot 2020-01-03 at 9 46 21 PM

### DIFF
--- a/internal/storage/conn.go
+++ b/internal/storage/conn.go
@@ -11,7 +11,7 @@ import (
 	"github.com/volatiletech/sqlboiler/boil"
 )
 
-const maxDBConnectAttempts = 10
+const maxDBConnectAttempts = 7
 
 // Handler implements the app database handler.
 type Handler interface {
@@ -63,14 +63,14 @@ func (c *Connection) Connect() error {
 	dsn := MakeDSN(c.params)
 	c.logger.LogF(monitor.F{"dsn": dsn}).Info("connecting to the DB")
 	var err error
-	var secondsToWait int
 	var db *sqlx.DB
 	for i := 0; i < maxDBConnectAttempts; i++ {
 		db, err = sqlx.Connect(c.dialect, dsn)
 		if err == nil {
 			break
 		}
-		secondsToWait = secondsToWait + i + 1
+		secondsToWait := i + 1
+		c.logger.Log().Warning("Database Connection Err: ", err)
 		c.logger.Log().Warningf("Attempt %d - could not connect to database...retry in %d seconds", i, secondsToWait)
 		time.Sleep(time.Duration(secondsToWait) * time.Second)
 	}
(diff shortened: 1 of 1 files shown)


internal/storage/conn.go
```diff
@@ -59,7 +62,19 @@ func InitConn(params ConnParams) *Connection {
...
-	db, err := sqlx.Connect(c.dialect, dsn)
+	var err error
+	var secondsToWait int
+	var db *sqlx.DB
+	for i := 0; i < maxDBConnectAttempts; i++ {
+		db, err = sqlx.Connect(c.dialect, dsn)
+		if err == nil {
+			break
+		}
+		secondsToWait = secondsToWait + i + 1
```
sayplastic: This is a neat idea. But having to wait up to 220 seconds for the database will create problems with k8s deployments where we don't want app in an uncertain state for that long. I think 30 seconds is a reasonable baseline.

internal/storage/conn.go
```diff
@@ -59,7 +62,19 @@ func InitConn(params ConnParams) *Connection {
...
+		c.logger.Log().Warningf("Attempt %d - could not connect to database...retry in %d seconds", i, secondsToWait)
```
sayplastic: Would help to include the details of the error.

internal/storage/conn.go
```diff
@@ -59,7 +62,19 @@ func InitConn(params ConnParams) *Connection {
...
-	db, err := sqlx.Connect(c.dialect, dsn)
+	var err error
+	var db *sqlx.DB
+	for i := 0; i < maxDBConnectAttempts; i++ {
+		db, err = sqlx.Connect(c.dialect, dsn)
+		if err == nil {
+			break
+		}
+		secondsToWait := i + 1
+		c.logger.Log().Warning("Database Connection Err: ", err)
```
sayplastic: A bit of a nitpick, but those two log entries are not guaranteed to reach the physical log stream together. Why not put it all in a single message?

tiger5226: No particular reason, besides I thought you would want them separated out because one is an error and one is an info statement. You never specified what you wanted, so I had to assume brother. so I went with a warning message because the error was causing a failure. I don't really have a strong opinion. So be specific and I will change it to whatever you want it to be.

sayplastic: A single `Warning` is fine here.

tiger5226:
@sayplastic How does coverage decrease if the function adjusted has a unit test covering it?

sayplastic:
> @sayplastic How does coverage decrease if the function adjusted has a unit test covering it?

Test coverage is not binary, you added a logical branch to the function that the existing test does not cover.

kauffj:
@tiger5226, @sayplastic says you can and should merge this

tiger5226:
All clear. I was just curious. I did not know if was that detailed of a check. Good to know.

PR #164, opened 2020-02-05
Bugfix/wallet loading

sayplastic:
This is to improve handling and alerting on various race conditions when wallet file is present on the filesystem but not loaded in memory. Depending on the SDK response, we attempt re-loading the wallet and re-issuing the request. 

What was not handled before is a race condition when wallet loading has already been performed by another parallel request to the API (as UI fires them very fast). In this case, SDK fails with an error saying `Wallet at path .+ is already loaded`, now we handle this error and transparently re-issue the request again instead of passing error back to the UI causing its failure.

### DIFF
--- a/app/proxy/client.go
+++ b/app/proxy/client.go
@@ -15,7 +15,7 @@ import (
 )
 
 const walletLoadRetries = 3
-const walletLoadRetryWait = time.Millisecond * 200
+const walletLoadRetryWait = time.Millisecond * 100
 
 var ClientLogger = monitor.NewModuleLogger("proxy_client")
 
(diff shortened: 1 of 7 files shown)


app/proxy/client.go
```diff
@@ -106,3 +111,13 @@ func (c *Client) isWalletNotLoaded(r *jsonrpc.RPCResponse) bool {
 	}
 	return false
 }
+
+func (c *Client) isWalletAlreadyLoaded(r *jsonrpc.RPCResponse) bool {
+	if r.Error != nil {
+		wErr := lbrynet.NewWalletError(0, errors.New(r.Error.Message))
+		if errors.As(wErr, &lbrynet.WalletAlreadyLoaded{}) {
```
lyoshenka: shouldn't this be `errors.Is`? i thought `errors.As` is for when you want to cast the error for use later

sayplastic: The way it's named is not exactly intuitive. Here's gist of how it works:

> In the simplest case, the errors.Is function behaves like a comparison to a sentinel error, and the errors.As function behaves like a type assertion.

PR #187, opened 2020-03-16
add caching to status response

nikooo777:


app/status/status.go
```diff
@@ -31,58 +31,73 @@ type ServerItem struct {
...
-	}
-	response := map[string]interface{}{
-		"timestamp":     fmt.Sprintf("%v", time.Now().UTC()),
-		"services":      services,
-		"general_state": StatusOK,
-	}
 	respStatus := http.StatusOK
-	failureDetected := false
+	var response *statusResponse
+	if cachedResponse != nil && lastUpdate.After(time.Now().Add(60*time.Second)) {
```
andybeletsky: Timeout should be defined as a constant at the top level

app/status/status.go
```diff
@@ -31,58 +31,73 @@ type ServerItem struct {
 }
 type ServerList []ServerItem
 
+type statusResponse map[string]interface{} {
+"timestamp":     fmt.Sprintf("%v", time.Now().UTC()),
```
andybeletsky: This results in a syntax error, check CI

andybeletsky:
Otherwise looks good to me

PR #189, opened 2020-03-23
Add buffer and time-to-start metrics

lyoshenka:


lyoshenka:
app-side PR https://github.com/lbryio/lbry-desktop/pull/3877

api/routes.go
```diff
@@ -31,6 +31,7 @@ func InstallRoutes(proxyService *proxy.ProxyService, r *mux.Router) {
 	internalRouter.Handle("/metrics", promhttp.Handler())
 	internalRouter.HandleFunc("/status", status.GetStatus)
 	internalRouter.HandleFunc("/whoami", status.WhoAMI)
+	internalRouter.HandleFunc("/ui_metric", status.TrackUIMetric)
```
andybeletsky: Since this is going to be accessed by clients, putting it in its own versioned namespace under `/api` makes sense (like `/api/v1/metric/ui`)

internal/status/status.go
```diff
@@ -118,3 +120,30 @@ func WhoAMI(w http.ResponseWriter, req *http.Request) {
 	respByte, _ := json.MarshalIndent(&details, "", "  ")
 	w.Write(respByte)
 }
+
+func TrackUIMetric(w http.ResponseWriter, req *http.Request) {
```
andybeletsky: As with any client facing code, using this without tests makes me uneasy...

internal/status/status.go
```diff
@@ -118,3 +120,30 @@ func WhoAMI(w http.ResponseWriter, req *http.Request) {
...
+	w.Header().Add("content-type", "application/json; charset=utf-8")
+	resp := make(map[string]string)
+
+	metricName := req.FormValue("name")
+	resp["name"] = metricName
+
+	switch metricName {
+	case "buffer":
+		metrics.UIBufferCount.Inc()
+		resp["status"] = "ok"
```
andybeletsky: I think it's okay to just return an empty response with 200 status code, will save a tiny bit of traffic too.

api/routes.go
andybeletsky: Should be POST-only to minimally prevent abuse.

internal/status/status.go
```diff
@@ -118,3 +120,30 @@ func WhoAMI(w http.ResponseWriter, req *http.Request) {
...
+	case "time_to_start":
+		t := cast.ToFloat64(req.FormValue("value"))
+		if t == 0 {
+			resp["error"] = "Time to start cannot be 0"
+		} else {
+			metrics.UITimeToStart.Observe(t)
+		}
+	default:
+		w.WriteHeader(http.StatusBadRequest)
+		resp["error"] = "Unknown metric"
```
andybeletsky: I think we want to log that.

andybeletsky:
👍

PR #197, opened 2020-04-13
SDK selection refactor

lyoshenka:
here's how I want to make wallet selection work:
- request comes into the proxy handler
- if user is not authenticated (aka does not have a verified email), they get a random sdk
- if they are authenticated but don't have an sdk assigned in the database, they get assigned the least-loaded sdk (and that's saved in the db)
- if they do have one assigned, they get the one they have
- all this logic happens in one place, either in a middleware or in a function that can be called to get the Caller. it should work the same way for proxy calls as for publishing (right now there are two slightly different paths)

i'm planning to drop the wallet_id column in the db, and most references to wallet ids in the code. sdk assignment will be done off user ids, and anytime wallet ids are needed, they'll just be generated from a template (right now that's `lbrytv-id.ID.wallet`)

other changes i made:
- dropped the Caller interface in favor of just using the real type. i get why it was necessary (for mocking the rpc client during testing), but we don't need that anymore because I ...
- rewrote tests to use a real jsonrpc server so we don't need to mock that anymore. and by real, i mean a server running in a separate thread that's created/destroyed for each test. look at internal/test/test.go if you're interested in how that works. it works equally well for mocking http servers

This is a continuation of the work in #195


app/proxy/client_test.go
```diff
@@ -108,29 +72,35 @@ func TestClientCallDoesNotReloadWalletIfAlreadyLoaded(t *testing.T) {
...
-
-	r, err := c.Call(q)
+	go func() {
+		srv.NextResponse <- test.ResToStr(t, jsonrpc.RPCResponse{
+			JSONRPC: "2.0",
+			Error: &jsonrpc.RPCError{
+				Message: "Couldn't find wallet: //",
+			},
+		})
+		srv.NextResponse <- "" // for the wallet_add call
```
andybeletsky, line 90: perhaps a matter of taste, but this would look more clean / self-explanatory:
EMPTY := ""

srv.NextResponse <- EMPTY

lyoshenka: i thought about that but didn't like that way. is `srv.NextResponse <- "{}"` better? or maybe an `SkipNext()` function on srv? i like SkipNext() less .

andybeletsky: I think SkipNext is more descriptive. Empty anything doesn't convey the intention and looks like a typo.

app/publish/publish.go
```diff
@@ -9,153 +9,95 @@ import (
...
 		return
 	}
+	defer func() {
+		if err := os.Remove(f.Name()); err != nil {
+			monitor.CaptureException(err, map[string]string{"file_path": f.Name()})
+		}
+	}()
 
-	response := h.Publisher.Publish(f.Name(), r.WalletID, []byte(r.FormValue(JSONRPCFieldName)))
+	w.Write(publish(authResult.SDKAddress, f.Name(), authResult.User().ID, []byte(r.FormValue(jsonRPCFieldName))))
```
andybeletsky, line 71: Many parentheses and calls to very different packages are happening here, I'd unfold it

app/proxy/client_test.go
```diff
@@ -6,137 +6,103 @@ import (
...
+	require.NoError(t, err)
+	q.WalletID = sdkrouter.WalletID(dummyUserID)
 
+	c := NewCaller(addr, dummyUserID)
+	r, err := c.callQueryWithRetry(q)
 	// err = json.Unmarshal(result, response)
 	require.NoError(t, err)
 	require.Nil(t, r.Error)
+
+	// TODO: check that wallet is actually reloaded? what is this test even testing?
```
andybeletsky, line 36: It is testing a situation where a wallet is existing but not currently loaded. Checking that SDK does not respond with an error, as it would if not for the logic in `callQueryWithRetry`, is enough.

app/proxy/const.go
```diff
@@ -140,12 +140,3 @@ var ignoreLog = []string{
 	MethodAccountBalance,
 	MethodStatus,
 }
-
-func shouldLog(method string) bool {
```
andybeletsky, line 144: Seems like I forgot to add an actual call for this, I would reinstate it, logs are swamped with `wallet_balance`.

server/server.go
```diff
@@ -19,63 +19,46 @@ var logger = monitor.NewModuleLogger("server")
...
+	}))
+
+	return &Server{
+		address:  address,
+		stopWait: 15 * time.Second,
+		stopChan: make(chan os.Signal),
+		listener: &http.Server{
+			Addr:    address,
+			Handler: r,
+			// Can't have WriteTimeout set for streaming endpoints
```
andybeletsky, line 45: We don't have those anymore so setting at least a write timeout would be a good idea.

internal/test/test_test.go
```diff
@@ -1,30 +1,45 @@
...
 	"net/http"
 	"testing"
 
 	ljsonrpc "github.com/lbryio/lbry.go/v2/extras/jsonrpc"
 
 	"github.com/stretchr/testify/assert"
+	"github.com/stretchr/testify/require"
 )
 
 func TestMockRPCServer(t *testing.T) {
```
andybeletsky, line 15: Probably should be renamed

internal/test/test.go
```diff
@@ -1,41 +1,157 @@
...
+
+	"github.com/lbryio/lbrytv/config"
+	"github.com/nsf/jsondiff"
+	"github.com/ybbus/jsonrpc"
 )
 
-type RequestData struct {
-	Request *http.Request
-	Body    string
+type MockServer struct {
```
andybeletsky, line 18: I like this idea.

internal/test/test.go
```diff
@@ -1,41 +1,157 @@
...
-	return httpServer, setNextResponse
+// JsonCompare compares two json strings.
+func jsonCompare(a, b []byte) (bool, string) {
+	opts := jsondiff.DefaultConsoleOptions()
+	diff, str := jsondiff.Compare(a, b, &opts)
+	return diff == jsondiff.FullMatch, str
+}
+
+// assert.Equal for JSON - more accurate comparison, pretty diff
+func AssertJsonEqual(t *testing.T, expected, actual interface{}, msgAndArgs ...interface{}) bool {
```
andybeletsky, line 111: This is going to save a lot of headache. Perhaps it would be a good idea to move out as a common package to lbry.go as it seems to suffer from "real world" test coverage because tests are not easy to write/run.

internal/status/status.go
```diff
@@ -44,53 +47,65 @@ type statusResponse map[string]interface{}
...
 		}
-		cachedResponse = response
+		cachedResponse = &response
 		lastUpdate = time.Now()
 	}
-	w.Header().Add("content-type", "application/json; charset=utf-8")
+
+	authResult := auth.FromRequest(req)
+	if authResult.Authenticated() {
+		response["user"] = map[string]interface{}{
```
andybeletsky, line 97: 👍 This is a long-awaited feature

internal/status/status.go
```diff
@@ -44,53 +47,65 @@ type statusResponse map[string]interface{}
 
 func GetStatus(w http.ResponseWriter, req *http.Request) {
```
andybeletsky, line 48: Now that this endpoint seems to be used officially in our code, it makes sense to promote it to a versioned api route (`/api/v1/status`) instead of `internal`.

internal/ip/ip.go
```diff
@@ -67,28 +67,28 @@ func IsPrivateSubnet(ipAddress net.IP) bool {
 }
 
 // GetIPAddressForRequest returns the real IP address of the request
```
andybeletsky, line 69: Docstring name should match

internal/ip/ip.go
```diff
@@ -67,28 +67,28 @@ func IsPrivateSubnet(ipAddress net.IP) bool {
...
-func GetIPAddressForRequest(r *http.Request) string {
+func AddressForRequest(r *http.Request) string {
 	for _, h := range []string{"X-Forwarded-For", "X-Real-Ip"} {
 		addresses := strings.Split(r.Header.Get(h), ",")
 		// march from right to left until we get a public address
 		// that will be the address right before our proxy.
 		for i := len(addresses) - 1; i >= 0; i-- {
-			ip := strings.TrimSpace(addresses[i])
+			addr := strings.TrimSpace(addresses[i])
```
andybeletsky, line 76: I improved this function in `master-stable` by adding tests. Perhaps the next step would be to move it out to a common package as right now there is another very similar copy of it in `internal-apis`.

cmd/serve.go
```diff
@@ -23,10 +22,7 @@ var rootCmd = &cobra.Command{
 		sdkRouter := sdkrouter.New(config.GetLbrynetServers())
 		go sdkRouter.WatchLoad()
 
-		s := server.NewServer(server.Options{
-			Address:      config.GetAddress(),
-			ProxyService: proxy.NewService(proxy.Opts{SDKRouter: sdkRouter}),
-		})
+		s := server.NewServer(config.GetAddress(), sdkRouter)
```
andybeletsky, line 25: 👍

andybeletsky:
Coverage decreased (-1.1%) to 63.347% though

app/auth/auth.go
```diff
@@ -0,0 +1,98 @@
...
+				res.authAttempted = true
+			}
+			next.ServeHTTP(w, r.Clone(context.WithValue(r.Context(), ContextKey, res)))
+		})
+	}
+}
+
+// wish i could make this non-exported, but then you can't create new providers outside the package
+// don't make this struct directly. instead use NewResult
+type Result struct {
```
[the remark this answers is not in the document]
lyoshenka: its an auth.Result - the result of authentication

i'm not sure the best way to do this. what i'd like is for it to either return a user, or an error, or neither. if neither, no authentication happened (aka there was no auth token). if error, then it happened and errors. if user, then it happened and was successful.

i'm open to ideas for how to do this better

app/auth/auth.go
```diff
@@ -0,0 +1,98 @@
...
+					logger.LogF(monitor.F{"ip": addr}).Debugf("error authenticating user")
+				}
+				res.authAttempted = true
+			}
+			next.ServeHTTP(w, r.Clone(context.WithValue(r.Context(), ContextKey, res)))
+		})
+	}
+}
+
+// wish i could make this non-exported, but then you can't create new providers outside the package
```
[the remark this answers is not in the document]
lyoshenka: the benefit is that it would be clearer that you can't have a user and an error at the same time

app/proxy/client_test.go
```diff
@@ -6,137 +6,103 @@ import (
...
+
+	srv := test.MockHTTPServer(nil)
+	defer srv.Close()
+
+	c := NewCaller(srv.URL, 0)
+	q, err := NewQuery(jsonrpc.NewRequest("wallet_balance"))
+	require.NoError(t, err)
+	q.WalletID = walletID
+
+	go func() {
```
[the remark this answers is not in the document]
lyoshenka: in the general case, you want the requests/responses to be sent via channels so you can get the request, do what you want with it, then send a response which might change based on the request. thats why test.MockServer uses channels. in this particular test, you're just sending responses without regard for the requests, so you can queue up a bunch of them and let them send. but the channel will block if you write to it without any reads coming out, so i did it in a goroutine.

in general i think having the mock server logic be in a goroutine makes a lot of sense. in production you have a client and a server running in parallel. they just happen to be running on the internet, not on localhost. this test mimics that.

app/proxy/processors.go
```diff
@@ -7,66 +7,56 @@ import (
...
 
 	ljsonrpc "github.com/lbryio/lbry.go/v2/extras/jsonrpc"
+
+	log "github.com/sirupsen/logrus"
 	"github.com/ybbus/jsonrpc"
 )
 
-func processQuery(query *jsonrpc.RPCRequest) (processedQuery *jsonrpc.RPCRequest, err error) {
-	processedQuery = query
+func postProcessResponse(response *jsonrpc.RPCResponse, query *jsonrpc.RPCRequest) error {
```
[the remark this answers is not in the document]
lyoshenka: there's a c.Preprocessor. you're right though, this logic needs work

internal/test/test.go
lyoshenka: not sure what you mean by real world test coverage. its using a library to do all the heavy lifting, then formatting it a bit nicer to match the assertion library we're using for the rest of the tests

app/proxy/const.go
lyoshenka: esp because that's our only way to see what's going on in the cluster right now

lyoshenka:
coverage actually went up in `app/` and `api/`. it went down in `internal/` though

lyoshenka:
a lot of unused functions in config.go. can i delete them?

andybeletsky:
> a lot of unused functions in config.go. can i delete them?

I think they're mostly from player code so yes.

app/auth/auth.go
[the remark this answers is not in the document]
andybeletsky: I would just return `nil, nil` for when there is neither an error nor a user

api/routes.go
```diff
@@ -1,51 +1,69 @@
...
-		panic(err)
-	}
+func InstallRoutes(r *mux.Router, sdkRouter *sdkrouter.Router) {
+	upHandler := &publish.Handler{UploadPath: config.GetPublishSourceDir()}
 
+	r.Use(recoveryHandler)
 	r.Use(methodTimer)
 
-	r.HandleFunc("/", Index)
+	r.HandleFunc("/", func(w http.ResponseWriter, req *http.Request) {
```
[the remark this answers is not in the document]
lyoshenka: what do you mean surprises in code discovery?

app/auth/auth.go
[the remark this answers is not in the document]
lyoshenka: but it also returns extra stuff, like sdk address. and might return more in the future. and it makes for a more complex interface when you're returning a bunch of values instead of just one

app/publish/handler_test.go
```diff
@@ -12,122 +14,176 @@ import (
...
-	authenticator := users.NewAuthenticator(&users.TestUserRetriever{WalletID: "UPldrAcc", Token: "uPldrToken"})
 	publisher := &DummyPublisher{}
-	pubHandler, err := NewUploadHandler(UploadOpts{Path: os.TempDir(), Publisher: publisher})
-	assert.NoError(t, err)
 
-	authenticator.Wrap(pubHandler.Handle).ServeHTTP(rr, req)
+	reqChan := test.ReqChan()
+	ts := test.MockHTTPServer(reqChan)
+	go func() {
+		req := <-reqChan
```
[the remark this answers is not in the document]
lyoshenka: discussing it here https://github.com/lbryio/lbrytv/pull/197#discussion_r411413657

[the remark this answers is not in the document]
lyoshenka: but also the Publisher interface probably should come back. i think it was a mistake to remove it

internal/ip/ip.go
lyoshenka: agreed

app/proxy/handlers.go
```diff
@@ -1,69 +1,78 @@
...
+		w.Write(NewJSONParseError(err).JSON())
+		return
+	}
 
-		if err != nil {
-			responses.JSONRPCError(w, err.Error(), ErrAuthFailed)
-			monitor.CaptureRequestError(err, r, w)
+	var userID int
+	var sdkAddress string
+	if MethodNeedsAuth(req.Method) {
```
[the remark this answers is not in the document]
lyoshenka: i agree. the way to fix this is to fix the caller/query divide, which i didn't have time for

app/proxy/client_test.go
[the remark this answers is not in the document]
andybeletsky: Guess I was just referring to the readability here more than the idea of using concurrency for this purpose. `mc.AddNextResponse()` is just a cleaner interface which can spawn the goroutine behind the scene.

config/config.go
```diff
@@ -37,21 +36,21 @@ func init() {
 	Config = GetConfig()
 }
 
-func GetConfig() *ConfigWrapper {
+func GetConfig() *configWrapper {
```
andybeletsky, line 39: Returning variables of private types from public functions is not a recommended practice

api/routes.go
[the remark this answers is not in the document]
andybeletsky: The kind of surprise when you read through `handlers.go` and realize that some of the handlers logic is actually in `routes.go`

app/wallet/wallet.go
```diff
@@ -0,0 +1,212 @@
...
+
+var logger = monitor.NewModuleLogger("wallet")
+
+func DisableLogger() { logger.Disable() } // for testing
+
+// TokenHeader is the name of HTTP header which is supplied by client and should contain internal-api auth_token.
+const TokenHeader = "X-Lbry-Auth-Token"
+const pgUniqueConstraintViolation = "23505"
+
+// Retrieve gets user by internal-apis auth token. If the user does not have a wallet yet, they
```
andybeletsky, line 29: docstring name doesn't match actual name

app/wallet/wallet.go
```diff
@@ -0,0 +1,212 @@
...
+// are assigned an SDK and a wallet is created for them on that SDK.
+func GetUserWithWallet(rt *sdkrouter.Router, internalAPIHost, token, metaRemoteIP string) (*models.User, error) {
```
andybeletsky, line 31: We are not returning wallets from this function, in fact, they are not mentioned anywhere in the function code, so the name is a bit misleading.

config/config.go
lyoshenka: can you link me to where this is not recommended? i see this in plenty of places and it makes sense to me

app/proxy/handlers.go
```diff
@@ -1,69 +1,78 @@
...
+	// We're in RPC-response-land from here on down
+	responses.AddJSONContentType(w)
 
-	q, err := NewQuery(body)
-	if err != nil || !methodInList(q.Method(), relaxedMethods) {
-		retriever := users.NewWalletService(rh.SDKRouter)
-		auth := users.NewAuthenticator(retriever)
-		walletID, err = auth.GetWalletID(r)
+	var req jsonrpc.RPCRequest
+	err = json.Unmarshal(body, &req)
```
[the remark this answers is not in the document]
lyoshenka: makes sense. i think the general solution is to make the handlers take a jsonrpc request, and return a jsonrpc response. this makes testing easy and lets you handle non-jsonrpc errors elsewhere. internal-apis does this

config/config.go
andybeletsky: I learned this from golint and [their reasoning](seems compelling), namely:

> can't put a *sample field in their own struct

Even in my not-so-long practice I distinctly remember being annoyed by this more than a few of times.

app/proxy/handlers.go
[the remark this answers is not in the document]
andybeletsky: Response formation is one part of it. Even more inconvenient is that all input has to be packed into http request data, headers etc.

PR #203, opened 2020-04-23
Recovery middleware

lyoshenka:
- copied in and improved our errors package. the one in lbrytv should replace the one in lbry.go
- new errors package has a method that's necessary for Sentry to show the correct stack trace for our errors
- merged and improved error handling middleware

future work:
- merge all our various go libraries into lbry.go so we don't have them littered around our repos

### DIFF
--- a/internal/errors/errors.go
+++ b/internal/errors/errors.go
@@ -5,13 +5,12 @@ import (
 	"fmt"
 	"reflect"
 	"runtime"
+
+	pkgerr "github.com/pkg/errors"
 )
 
 // The maximum number of stack frames on any error.
 const maxStackDepth = 50
-
-// interop with pkg/errors
-type causer interface{ Cause() error }
 
 // traced is an error with an attached stack trace
 type traced struct {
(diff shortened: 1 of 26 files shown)


internal/errors/errors.go
```diff
@@ -0,0 +1,162 @@
...
+	return reflect.TypeOf(e.err).String()
+}
+
+//// New replaces builtin errors.New()
+//func New(text string) error {
+//	return Err(text)
+//}
+
+// Err returns an error with stack trace
+func Err(e interface{}, fmtParams ...interface{}) error {
```
andybeletsky, line 79: Why not call it Wrap? For drop-in compatibility with `pkg/errors`, if not to avoid `errors.Err(err)`

internal/errors/errors.go
```diff
@@ -0,0 +1,162 @@
+package errors
```
andybeletsky, line 1: A package as widely used as this would certainly benefit from some unit tests

internal/errors/errors.go
lyoshenka: it doesn't just wrap errors, but creates new errors as well. You can give it a string or an error. And sometimes it doesn't wrap or create an error, just returns the existing error if it's already one of ours.

I'm open to another name. I couldn't think of another one.

internal/errors/errors.go
lyoshenka: :100:

lyoshenka: at the same time, the functions are so simple that i'm not really sure how much tests would add here

lyoshenka: i guess i could at least test interop with pkg/errors and sentry

PR #210, opened 2020-04-27
drop wallet_id column on users

lyoshenka:


internal/storage/migrations/0004_drop_wallet_id.sql
```diff
@@ -0,0 +1,8 @@
+-- +migrate Up
+
+UPDATE users SET lbrynet_server_id = (id%30)+1 WHERE lbrynet_server_id IS NULL;
```
andybeletsky, line 3: This will make it work on dev as well:

`UPDATE users SET lbrynet_server_id = (id % (SELECT COUNT(id) from lbrynet_servers))+1 WHERE lbrynet_server_id IS NULL;`

andybeletsky:
otherwise lgtm

lyoshenka:
merged

PR #213, opened 2020-04-30
Enable non-wallet-specific SDK methods to accept wallet_id

andybeletsky:


app/proxy/handlers.go
```diff
@@ -44,7 +44,7 @@ func Handle(w http.ResponseWriter, r *http.Request) {
 
 	var userID int
 	var sdkAddress string
-	if MethodNeedsAuth(req.Method) {
+	if MethodRequiresWallet(req.Method) {
```
lyoshenka, line 47: :+1:

internal/test/test.go
```diff
@@ -90,6 +90,16 @@ func ResToStr(t *testing.T, res jsonrpc.RPCResponse) string {
 	return string(r)
 }
 
+// StrToRes creates an RPCResponse from a supplied string
+func StrToRes(t *testing.T, res string) jsonrpc.RPCResponse {
```
lyoshenka, line 94: prolly i should have just completed the set when i wrote the other three...

PR #216, opened 2020-05-07
Add wallet last access tracking and command to unload unused wallets

lyoshenka:
**Don't merge yet**. Let's test on ops cluster before pushing to prod.

This builds on #215


app/wallet/accesstracker/accesstracker_test.go
```diff
@@ -0,0 +1,98 @@
...
+	assert.True(t, u.WalletAccessedAt.Valid)
+
+	// touch user
+	Touch(boil.GetDB(), u.ID)
+
+	// reload, check that access time has been updated
+	err = u.ReloadG()
+	require.NoError(t, err)
+	assert.True(t, u.WalletAccessedAt.Valid)
+	assert.True(t, TimeNow().Add(-1*time.Minute).Before(u.WalletAccessedAt.Time))
```
andybeletsky, line 65: Why not compare it to the initial u.WalletAccessedAt?

internal/storage/migrations/0005_wallet_access.sql
```diff
@@ -0,0 +1,8 @@
+-- +migrate Up
+
+ALTER TABLE users ADD COLUMN "wallet_accessed_at" timestamp DEFAULT NULL;
```
andybeletsky, line 3: As there's not much else being accessed, I would simply call it `seen_at`

app/wallet/accesstracker/accesstracker_test.go
```diff
@@ -0,0 +1,98 @@
...
+}
+
+func TestUnload(t *testing.T) {
+	reqChan := test.ReqChan()
+	ts := test.MockHTTPServer(reqChan)
+	defer ts.Close()
+	ts.NextResponse <- `{"id":1}` // for the wallet.Unload call
```
andybeletsky, line 72: Not entirely clear that we're mocking wallet_remove response here. But perhaps we shouldn't be mocking it at all? Would be good to check that it actually unloads the wallet and that it can be loaded later.

app/wallet/accesstracker/accesstracker_test.go
```diff
@@ -0,0 +1,98 @@
+package accesstracker
```
andybeletsky, line 1: I propose just `tracker`, as per Effective Go:

> package name should be good: short, concise, evocative. By convention, packages are given lower case, single-word names; there should be no need for underscores or mixedCaps

app/proxy/proxy.go
```diff
@@ -79,7 +80,9 @@ func Handle(w http.ResponseWriter, r *http.Request) {
 	if cache.IsOnRequest(r) {
 		qCache = cache.FromRequest(r)
 	}
-	c := query.NewCallerWithCache(sdkAddress, userID, qCache)
+	c := query.NewCaller(sdkAddress, userID)
+	c.Cache = qCache
+	c.DB = boil.GetDB()
```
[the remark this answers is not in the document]
lyoshenka: they are not always required (e.g. in tests). i could put them in the initializer, but then you have to pass `nil` all the time. i'm open to it though. it would make the dependencies more explicit.

cmd/unload_wallets.go
```diff
@@ -0,0 +1,50 @@
...
+	log "github.com/sirupsen/logrus"
+	"github.com/spf13/cobra"
+	"github.com/volatiletech/sqlboiler/boil"
+)
+
+func init() {
+	rootCmd.AddCommand(unloadWallets)
+}
+
+var unloadWallets = &cobra.Command{
```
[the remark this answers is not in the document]
lyoshenka: :+1:

app/wallet/accesstracker/accesstracker.go
```diff
@@ -0,0 +1,91 @@
...
+		if err != nil {
+			l.Error(err)
+			continue
+		}
+
+		// only mark wallet unloaded if it hasn't been touched since we ran the query
+		// otherwise it may never be unloaded
+		q := fmt.Sprintf(`UPDATE "%s" SET "%s" = NULL WHERE "%s" = $1 AND "%s" = $2`,
+			models.TableNames.Users,
+			models.UserColumns.WalletAccessedAt,
```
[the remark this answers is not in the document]
lyoshenka: if we don't clear it, we'll try to unload that wallet again the next time this command runs

internal/storage/migrations/0005_wallet_access.sql
lyoshenka: but its not when the user was last seen. they could have been seen a bunch of times, but if they didn't call a query that uses the wallet, it won't be updated

app/wallet/accesstracker/accesstracker_test.go
lyoshenka: im mocking it because i'm testing whether it makes the call, not whether the sdk does what its supposed to do. that can probably be another test if you like

app/wallet/accesstracker/accesstracker_test.go
lyoshenka: :+1: i didnt like the name i chose

app/publish/publish.go
```diff
@@ -81,7 +82,9 @@ func (h Handler) Handle(w http.ResponseWriter, r *http.Request) {
 }
 
 func publish(sdkAddress, filename string, userID int, qCache cache.QueryCache, rawQuery []byte) []byte {
-	c := query.NewCallerWithCache(sdkAddress, userID, qCache)
+	c := query.NewCaller(sdkAddress, userID)
+	c.Cache = qCache
+	c.DB = boil.GetDB()
```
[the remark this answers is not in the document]
lyoshenka: the only way i see of doing this in a middleware is to touch the user anytime they provide their access token, regardless of whether they used the wallet or not. 

i guess i'm fine with this approach, combined with changing the column to `seen_at`. my original intention was to unload unused wallets. this would change that slightly, to unloading wallets of inactive users. that's close enough to the same thing for our purposes that i'm ok with this in exchange for reduced coupling.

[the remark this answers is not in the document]
andybeletsky: There's currently no scenario where a user would be on the website and not use his wallet. If not for the fact resolve calls include `wallet_id` now, `wallet_balance` is called every second.

app/wallet/tracker/tracker.go
```diff
@@ -0,0 +1,119 @@
...
+
+func GetLogger() monitor.ModuleLogger { return logger } // for testing
+
+var utcLocation, _ = time.LoadLocation("UTC")
+
+// TimeNow returns the current time in UTC. The only way to set the local timezone in Go is to
+// set the TZ env var. Otherwise all times are created using the local timezone of your OS. This
+// can screw up your time comparisons if you're not careful. Best practice: always use UTC
+func TimeNow() time.Time {
+	return time.Now().In(utcLocation)
```
andybeletsky, line 31: `Time.Now().UTC()`

internal/storage/migrations/0005_wallet_access.sql
```diff
@@ -0,0 +1,8 @@
+-- +migrate Up
+
+ALTER TABLE users ADD COLUMN "last_seen_at" timestamp DEFAULT NULL;
```
andybeletsky, line 3: Gonna need an index on this column.

cmd/unload_wallets.go
```diff
@@ -0,0 +1,37 @@
...
+		min, err := strconv.Atoi(args[1])
+		if err != nil {
+			log.Error(args[1] + " is not an integer")
+			os.Exit(1)
+		}
+
+		unloadOlderThan := time.Duration(min) * time.Minute
+		_, err = tracker.Unload(boil.GetDB(), unloadOlderThan)
+		if err != nil {
+			log.Error(err)
```
andybeletsky, line 33: It would be great if this was sent to sentry. Otherwise will only be visible if someone looks in the `CronJob` logs for it.

app/wallet/tracker/tracker.go
lyoshenka: :+1:

cmd/unload_wallets.go
lyoshenka: :+1:

internal/storage/migrations/0005_wallet_access.sql
lyoshenka: done

lyoshenka:
merged

PR #218, opened 2020-05-13
Implement paid token framework

andybeletsky:
This is a module for creating and validating paid stream tokens. This package will be used both on the API side and in the player code.

- `paid.CreateToken` creates a signed JWT on the lbrytv API side and supplies to the client via patched JSON-RPC `get` method (to be implemented)
- `paid.CanPlayStream` is called by lbrytv-player to validate incoming tokens

TODO:

- ~Figure out how to keep keys in database so we don’t regenerate them on each get~  <-- we are regenerating them on each `get`, there doesn't seem to be a downside to that
- ~Patch get method~


app/paid/paid.go
```diff
@@ -0,0 +1,122 @@
...
+	privKey         *rsa.PrivateKey
+	pubKeyMarshaled []byte
+}
+
+var km *keyManager
+
+// InitPrivateKey loads a private key from `file` for later token signing and derived pubkey distribution.
+// Should be generated as follows:
+// $ openssl genrsa -out privateKey.pem 2048
+func InitPrivateKey(file string) error {
```
lyoshenka, line 39: i recommend using io.Reader instead of a filename here. makes interface more flexible and testing easier

app/paid/paid.go
```diff
@@ -0,0 +1,122 @@
...
+	TxID   string `json:"txid"`
+	jwt.StandardClaims
+}
+
+type keyManager struct {
+	privKey         *rsa.PrivateKey
+	pubKeyMarshaled []byte
+}
+
+var km *keyManager
```
lyoshenka, line 34: would it be better to do dependency injection instead of a global var?

app/paid/paid_test.go
```diff
@@ -0,0 +1,74 @@
...
+	"github.com/stretchr/testify/require"
+)
+
+var testSDHash = "bea4d30a1868a00e98297cfe8cdefc1be6c141b54bea3b7c95b34a66786c22ab4e9f35ae19aa453b3630e76afbd24fe2"
+var testTxID = "e5a6a9ef4433b8868cebb770d1d2244eed410eb7503a9ffa82dce40aa62bbae7"
+
+func generateKeyFile() (string, error) {
+	keyFilename := fmt.Sprintf("test_private_key_%v.pem", time.Now().Unix())
+	cmd := fmt.Sprintf("openssl genrsa -out %s 2048", keyFilename)
+	if _, err := exec.Command("bash", "-c", cmd).Output(); err != nil {
```
lyoshenka, line 22: feels sketchy. id rather do this inside our code

app/paid/paid_test.go
```diff
@@ -0,0 +1,74 @@
...
+	}
+
+	pubKM = &pubKeyManager{}
+	err = pubKM.loadFromBytes(km.PublicKeyBytes())
+	if err != nil {
+		panic(err)
+	}
+
+	code := m.Run()
+	os.Remove(keyFile) // defers do not run after os.Exit()
```
lyoshenka, line 46: this is another thing `io.Reader` solves

app/paid/paid_test.go
lyoshenka: looks like you use `rsa.GenerateKey(rand.Reader, 2048)` in other tests. id do the same here

app/paid/validator.go
```diff
@@ -0,0 +1,83 @@
...
+var pubKM *pubKeyManager
+
+// InitPubKey should be called with pubkey url as an argument before CanPlayStream can be called
+func InitPubKey(url string) error {
+	pubKM = &pubKeyManager{url: url}
+	err := pubKM.Download()
+	if err != nil {
+		return err
+	}
+	return nil
```
lyoshenka, line 29: this could be simply `return pubMK.Download()`

app/paid/validator.go
```diff
@@ -0,0 +1,83 @@
...
+
+	"github.com/dgrijalva/jwt-go"
+)
+
+type pubKeyManager struct {
+	url string
+	key *rsa.PublicKey
+}
+
+var pubKM *pubKeyManager
```
lyoshenka, line 20: something about these singleton instances feels wrong. all its really doing is storing an rsa key (the `url` is only used in one place to download the key). 

this whole file could be replaced by two functions: `Load([]byte) *rsa.PublicKey` to load the key from bytes, and `Validate(key *rsa.PublicKey, token string) (*StreamToken, error)` to validate the token using the key. then you load the key elsewhere (maybe in InstallRoutes or some other setup function) and put it on the requests that need it

app/paid/paid_test.go
andybeletsky: I wanted this part to be as close to real world scenario as possible — it's just easier to generate the key file with openssl during deployment

app/paid/validator.go
andybeletsky: But the logic of downloading the key from a certain url still has to reside somewhere. Doesn't feel like routes file or any of the initialization files are great for this purpose

andybeletsky:
@lyoshenka I added `get` patching, could you please review?

app/paid/validator.go
andybeletsky: I agree that they don't feel quite right though.

andybeletsky: I think I'm actually at peace with those. They are singletons containing some persistent state that cannot be stored in any other way and some logic about that state that only gets called by two publicly exposed functions. I think this is the easiest way to go about it in a way that would make for easy re-use of this in lbrytv-player code 

:change_my_mind.jpg:

lyoshenka: is it weird that the pubkey manager is downloading a key from the internet? feels like it shouldn't be responsible for that. instead of passing a URL to Init(), i'd pass is the data (or an io.Reader).

app/query/caller_test.go
```diff
@@ -401,11 +413,272 @@ func TestSDKMethodStatus(t *testing.T) {
...
 
-func parseRawResponse(t *testing.T, rawCallResponse []byte, v interface{}) {
-	assert.NotNil(t, rawCallResponse)
-	var res jsonrpc.RPCResponse
-	err := json.Unmarshal(rawCallResponse, &res)
+func TestCaller_GetPaidCannotPurchase(t *testing.T) {
+	dummyUserID := rand.Intn(99999)
+	srvAddress := test.RandServerAddress(t)
+	uri := "lbry://@specialoperationstest#3/iOS-13-AdobeXD#9"
+	err := wallet.Create(srvAddress, dummyUserID)
```
lyoshenka, line 420: require no error here

app/query/caller_test.go
```diff
@@ -401,11 +413,272 @@ func TestSDKMethodStatus(t *testing.T) {
...
+
+	request := jsonrpc.NewRequest(MethodGet, map[string]interface{}{"uri": uri})
+	rawCallResponse := NewCaller(srvAddress, dummyUserID).Call(request)
+
+	var errorResponse jsonrpc.RPCResponse
+	err = json.Unmarshal(rawCallResponse, &errorResponse)
```
lyoshenka, line 426: look at `test.StrToRes` for a shortcut for this

app/query/caller_test.go
```diff
@@ -401,11 +413,272 @@ func TestSDKMethodStatus(t *testing.T) {
...
 	require.NoError(t, err)
-	err = res.GetObject(v)
+	fmt.Println(string(rawCallResponse))
+	assert.Equal(t, "Not enough funds to cover this transaction.", errorResponse.Result.(map[string]interface{})["error"])
+}
+
+func TestCaller_GetPaidPurchased(t *testing.T) {
+	config.Override("BaseContentURL", "https://cdn.lbryplayer.xyz/api/v2/streams/")
```
lyoshenka, line 433: every time i see a `config.Override()` in a test, it makes me think that that value should be passed in to a `NewX()` func during initialization rather gotten directly from the global config. do you feel the same way?

app/query/caller_test.go
```diff
@@ -401,11 +413,272 @@ func TestSDKMethodStatus(t *testing.T) {
...
+}
+
+func TestCaller_GetFree(t *testing.T) {
+	config.Override("BaseContentURL", "https://cdn.lbryplayer.xyz/api/v2/streams/")
+	defer config.RestoreOverridden()
+
+	uri := "lbry://what"
+
+	dummyUserID := 123321
+	srv := test.MockHTTPServer(nil)
```
lyoshenka, line 607: defer srv.Close()

app/query/caller_test.go
andybeletsky: It does feel like that. But I guess compromises have to be made. This variable is just used once, to promote it to an initializer parameter feels wasteful.

app/paid/validator.go
andybeletsky: I see your point. Agreed

app/paid/paid_test.go
andybeletsky: went with []byte

andybeletsky:
`paid` package has been [moved over to lbrytv-player](https://github.com/lbryio/lbrytv-player/pull/14).

app/query/processors.go
```diff
@@ -1,97 +1,122 @@
...
 
 	ljsonrpc "github.com/lbryio/lbry.go/v2/extras/jsonrpc"
 
-	log "github.com/sirupsen/logrus"
 	"github.com/ybbus/jsonrpc"
 )
 
-func postProcessResponse(response *jsonrpc.RPCResponse, query *jsonrpc.RPCRequest) error {
-	switch query.Method {
+type PurchaseReceipt struct {
```
lyoshenka, line 14: no need to make this exported. I always err on the side of leaving everything unexported. Keeps the package's API small. If you need to export it later, its very easy. Going the other way is harder.

andybeletsky: As someone who did Python for too long, this feels a bit foreign. Is there a good reason to err on the side of privateness? 

I remember mentioning in another discussion that I was inconvenienced a few times by stumbling into variables that were private. But I don't recall ever saying to myself: damn, I wish this `type` was made private by the package author.

internal/test/test.go
```diff
@@ -59,7 +59,7 @@ func MockHTTPServer(requestChan chan *Request) *mockServer {
 // ReqChan makes a channel for reading received requests one by one.
 // Use it in conjunction with MockHTTPServer
 func ReqChan() chan *Request {
-	return make(chan *Request, 1)
+	return make(chan *Request, 999)
```
andybeletsky, line 62: @lyoshenka any downsides to this? it was blocking for me when tested code was calling fake server more times than expected

app/query/processors.go
lyoshenka: here's how i see it: think of every Go package (including all the sub-packages in lbrytv) as it's own library, and every exported thing as part of that library's API. the API is a promise to other packages. if you want to add to the API, it's easy. if you want to change the API or drop things, you're now breaking your promise to any package that uses your package.

with those incentives, you want to promise as little as possible. if another package needs something from you, you can export an unexported object to add to your API. if you export too much, you'll have to consider more possibilities when you do want to change something. 

related: only exporting what's useful to other packages reduces the cognitive load on those packages. pretend i'm an external developer. if i see `query.PurchaseReceipt` exported, that's an [affordance](https://www.interaction-design.org/literature/topics/affordances). i may say "oh cool, a thing i can use for something. what does it do?". but in reality there's nothing for me to do with it. its not useful to me.

this is related to the logic behind deleting code that's not used, and also related to the principle of least privilege (though not in a security sense).

lyoshenka: speaking of Python, I don't like "private-by-convention" at all. i get the logic for it and the appeal. treating other devs as "responsible adults who play nice together" is reasonable. it also increases the range of possibilities and opens the door to mistakes. 

for more on this, i highly recommend Rich Hickey's [Simple Made Easy](https://github.com/matthiasn/talk-transcripts/blob/master/Hickey_Rich/SimpleMadeEasy.md). his idea of complecting as the opposite of simplicity is very influential to me

internal/test/test.go
lyoshenka: i don't really like the way we use channels here, even though im the one who wrote this. bumping the buffer to 999 just hides the problem, and i don't like hiding problems. i don't have a good solution yet, so this is prolly fine for now.

andybeletsky:
@lyoshenka would appreciate a review of the latest caller refactoring. I added a unified pre-flight hook mechanism instead of numerous function calls.

PR #228, opened 2020-05-29
Improvement/status

andybeletsky:
I added a status endpoint V2:

- calls the server our user is on instead of polling all 40 servers (for unauthenticated this will be a random server every time since they don't get assigned a specific one but that's acceptable IMO)
- calls `reverse` instead of `status`
- doesn't poll players

I also have removed lbrynet servers from the local config file, they're now coming from the database, i.e. closer to how it's done in production.

### DIFF
--- a/app/query/caller.go
+++ b/app/query/caller.go
@@ -88,6 +88,10 @@ func (c *Caller) addDefaultHooks() {
 // Call method forwards a JSON-RPC request to the lbrynet server.
 // It returns a response that is ready to be sent back to the JSON-RPC client as is.
 func (c *Caller) Call(req *jsonrpc.RPCRequest) (*jsonrpc.RPCResponse, error) {
+	if c.endpoint == "" {
+		return nil, errors.Err("cannot call blank endpoint")
+	}
+
 	walletID := ""
 	if c.userID != 0 {
 		walletID = sdkrouter.WalletID(c.userID)
--- a/app/query/caller.go
+++ b/app/query/caller.go
@@ -58,7 +58,7 @@ func NewCaller(endpoint string, userID int) *Caller {
 						KeepAlive: 120 * time.Second,
 					}).Dial,
 					TLSHandshakeTimeout:   30 * time.Second,
-					ResponseHeaderTimeout: 300 * time.Second,
+					ResponseHeaderTimeout: 600 * time.Second,
 					ExpectContinueTimeout: 1 * time.Second,
 				},
 			},
@@ -149,6 +149,7 @@ func (c *Caller) callQueryWithRetry(q *Query) (*jsonrpc.RPCResponse, error) {
 		// Generally a HTTP transport failure (connect error etc)
 		if err != nil {
 			logger.Log().Errorf("error sending query to %v: %v", c.endpoint, err)
+			metrics.ProxyCallFailedDurations.WithLabelValues(q.Method(), c.endpoint, metrics.FailureKindNet).Observe(duration)
 			return nil, errors.Err(err)
 		}
 
@@ -188,7 +189,7 @@ func (c *Caller) callQueryWithRetry(q *Query) (*jsonrpc.RPCResponse, error) {
 	if err != nil || (r != nil && r.Error != nil) {
 		logFields["response"] = r.Error
 		logger.WithFields(logFields).Error("rpc call error")
-		metrics.ProxyCallFailedDurations.WithLabelValues(q.Method(), c.endpoint).Observe(duration)
+		metrics.ProxyCallFailedDurations.WithLabelValues(q.Method(), c.endpoint, metrics.FailureKindRPC).Observe(duration)
 	} else {
 		if config.ShouldLogResponses() {
 			logFields["response"] = r
--- a/internal/status/status.go
+++ b/internal/status/status.go
@@ -45,9 +45,9 @@ const (
 )
 
 type ServerItem struct {
-	Address string `json:"address"`
-	Status  string `json:"status"`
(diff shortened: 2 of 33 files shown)


andybeletsky:
That feeling when you add two and a half tests, yet the coverage decreases.

app/query/caller.go
```diff
@@ -88,6 +88,10 @@ func (c *Caller) addDefaultHooks() {
 // Call method forwards a JSON-RPC request to the lbrynet server.
 // It returns a response that is ready to be sent back to the JSON-RPC client as is.
 func (c *Caller) Call(req *jsonrpc.RPCRequest) (*jsonrpc.RPCResponse, error) {
+	if c.endpoint == "" {
+		return nil, errors.Err("cannot call blank endpoint")
```
lyoshenka, line 92: :+1:

internal/status/status.go
```diff
@@ -91,11 +96,79 @@ func GetStatus(w http.ResponseWriter, req *http.Request) {
...
+	w.Write(respByte)
+}
+
+func GetStatusV2(w http.ResponseWriter, r *http.Request) {
+	respStatus := http.StatusOK
+	var response statusResponse
+
+	if cachedResponse != nil && lastUpdate.After(time.Now().Add(statusCacheValidity)) {
+		copier.Copy(&response, cachedResponse)
+	} else {
```
lyoshenka, line 114: instead of this giant else, i would put a `return` after `copier.Copy(&response, cachedResponse)`. this way you avoid an extra level of nested braces. its the recommended style in go: https://golang.org/doc/effective_go.html#if

internal/status/status.go
```diff
@@ -91,11 +96,79 @@ func GetStatus(w http.ResponseWriter, req *http.Request) {
...
+			userID     int
+			qCache     cache.QueryCache
+			sdkAddress string
+		)
+		rt := sdkrouter.New(config.GetLbrynetServers())
+		user, err := auth.FromRequest(r)
+		if err != nil {
+			sdkAddress = rt.RandomServer().Address
+		} else {
+			sdkAddress = auth.SDKAddress(user)
```
lyoshenka, line 135: i believe its possible that there's no error, but also there's no user. this happens if the auth token is provided but doesn't match any user in internal-apis

if you find this behavior confusing, we can have it error in that case as well. i was not sure which option is better when i wrote this so i'd love to hear your thoughts

andybeletsky:
Tried to remove sdk servers from the config file but I was defeated. Still some quirks to work out there.
I moved/renamed `auth.SDKAddress` function to `sdkrouter` package, felt like it didn't belong to `auth`.

internal/status/status.go
```diff
@@ -44,9 +45,9 @@ const (
 )
 
 type ServerItem struct {
```
lyoshenka, line 47: does ServerItem need to be exported?

internal/status/status_test.go
```diff
@@ -0,0 +1,147 @@
...
+		Connection: dbConfig.Connection,
+		DBName:     dbConfig.DBName,
+		Options:    dbConfig.Options,
+	}
+	c, connCleanup := storage.CreateTestConn(params)
+	c.SetDefaultConnection()
+
+	defer connCleanup()
+
+	os.Exit(m.Run())
```
lyoshenka, line 39: this doesn't work. `os.Exit()` will kill the process immediately without running any deferred functions. take a look at your postgres - i bet you have a million temp databases.

look at how i refactored the other tests that use CreateTestConn. you should do it that way

internal/status/status_test.go
```diff
@@ -0,0 +1,147 @@
...
+	respBody, err := ioutil.ReadAll(response.Body)
+	require.NoError(t, err)
+
+	assert.Equal(t, http.StatusOK, response.StatusCode)
+	var respStatus statusResponse
+	err = json.Unmarshal(respBody, &respStatus)
+	require.NoError(t, err)
+
+	assert.Equal(t, statusFailing, respStatus["general_state"])
+	lbrynetStatus := respStatus["services"].(map[string]interface{})["lbrynet"].(map[string]interface{})
```
lyoshenka, line 85: :woozy_face:

internal/status/status_test.go
```diff
@@ -0,0 +1,147 @@
...
+		"user_id": 123,
+		"has_verified_email": true
+  	}
+}`
+
+	rr := httptest.NewRecorder()
+	r, _ := http.NewRequest("GET", "", nil)
+	r.Header.Add(wallet.TokenHeader, "anystringwilldo")
+	rt := sdkrouter.New(config.GetLbrynetServers())
+	handler := sdkrouter.Middleware(rt)(auth.Middleware(auth.NewIAPIProvider(rt, ts.URL))(http.HandlerFunc(GetStatusV2)))
```
lyoshenka, line 115: also :woozy_face:, and i think you pointed this out to me in the past. maybe we could have a `middleware.Apply` function that takes a handler and a slice of middlewares and applies them? would be more readable

internal/status/status.go
andybeletsky: I thought it does for json marshaller to work properly but seems like that's not true.

lyoshenka: The fields of the struct have to be exported, but the struct itself does not.

internal/status/status_test.go
andybeletsky: I added a few helpers to `auth` and moved out middleware tools into their own package. Feels like parts of tests involving auth machinery initialization are more readable now.

PR #231, opened 2020-06-02
Postgres tx fix

lyoshenka:
Things to note:

1. `os.Exit()` doens't run deferred functions so we weren't cleaning up our test databases. This fixes that, but keep in mind that if you run tests inside Goland and you stop the test partway, it'll still leave the db around.

2. if a query errors inside a postgres transaction, the transaction is aborted and you have to roll it back and start over from scratch. i added code to do that once automatically. the concrete use case for us is if you try to insert a user in a tx and the user exists, you can't fetch that user inside the same tx

### DIFF
--- a/app/wallet/wallet_test.go
+++ b/app/wallet/wallet_test.go
@@ -1,6 +1,8 @@
 package wallet
 
 import (
+	"context"
+	"database/sql"
 	"fmt"
 	"io/ioutil"
 	"math/rand"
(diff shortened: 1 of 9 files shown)


app/wallet/wallet_test.go
```diff
@@ -361,3 +365,49 @@ func TestCreateWalletLoadWallet(t *testing.T) {
...
+	err = inTx(context.Background(), storage.Conn.DB.DB, func(tx *sql.Tx) error {
+		mockExecutor.ex = tx
+		_, err := getOrCreateLocalUser(mockExecutor, id, logger.Log())
+		return err
+	})
+
+	assert.NoError(t, err)
+}
+
+type firstQueryNoResults struct {
```
lyoshenka, line 392: @andybeletsky idk how to feel about this

PR #247, opened 2020-07-10
Initial commit for collector app

andybeletsky:
This is an attempt to implement collector app as a part of monorepo. Here is a list of key parts of it:

- `./apps` — where separate apps are supposed to live
- `./apps/collector` — contains collector app, its migrations, some configs and other resource files
- `./apps/collector/openapi.yml` — openapi spec for collector API, used primarily for validation
- `./apps/environment` — a common space for initialization procedures for each of the future "apps"
- `./pkg/app` — a main initializer and http server for an "app"


apps/collector/collector/event_types.go
```diff
@@ -0,0 +1,20 @@
...
+	StreamDuration *int   `json:"stream_duration,omitempty"`
+	StreamBitrate  *int   `json:"stream_bitrate,omitempty"`
+}
+
+type PlaybackStart struct {
+	// media bitrate, bits/sec (if known)
+	StreamBitrate *int `json:"stream_bitrate,omitempty"`
+	// stream duration, seconds
+	StreamDuration *int `json:"stream_duration,omitempty"`
+	// time before playback started, milliseconds
```
lyoshenka, line 16: these comments make me think that the names of the fields are not clear enough. the name should tell you exactly what the thing is.

also, making some fields be seconds and some be milliseconds is a recipe for mistakes. pick one or the other and make all times be that.

apps/collector/openapi.yml
```diff
@@ -0,0 +1,97 @@
+openapi: "3.0.0"
```
lyoshenka, line 1: ive been looking forward to using openapi :+1:

collector.yml
```diff
@@ -0,0 +1,4 @@
+DatabaseDSN: postgres://collector:ZnPfJ!4vTb-nURv8M!JkkuAcjYLYbkiX@localhost:5433
```
lyoshenka, line 1: should these passwords be in here and in sqlboiler.toml?

lyoshenka: or are these just for testing/docker?

apps/collector/collector/event_types.go
andybeletsky: The comments are auto-generated from the openapi spec. Agree on the units.

collector.yml
andybeletsky: These are for local testing. Production will be completely different

PR #278, opened 2020-10-06
E2e metrics

andybeletsky:
This adds measuring middleware so we can measure duration for the whole lifecycle of HTTP request, including time spent in other middlewares.

### DIFF
--- a/internal/metrics/middleware.go
+++ b/internal/metrics/middleware.go
@@ -13,7 +13,9 @@ type key int
 
 const timerContextKey key = iota
 
-// Chain chains multiple middleware together
+// Measure middleware starts a timer whenever a request is performed.
+// Note that it doesn't catch any metrics by itself,
+// HTTP handlers are expected to add their own by calling AddObserver.
 func Measure() mux.MiddlewareFunc {
 	return func(next http.Handler) http.Handler {
 		return http.HandlerFunc(func(w http.ResponseWriter, r *http.Request) {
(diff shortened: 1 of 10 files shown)


internal/metrics/middleware.go
```diff
@@ -0,0 +1,53 @@
...
+
+type key int
+
+const timerContextKey key = iota
+
+// Measure middleware starts a timer whenever a request is performed.
+// It should be added as first in the chain of middlewares.
+// Note that it doesn't catch any metrics by itself,
+// HTTP handlers are expected to add their own by calling AddObserver.
+func Measure() mux.MiddlewareFunc {
```
lyoshenka, line 20: lets call this `Middleware()` to be consistent with the other middleware we have

internal/metrics/middleware.go
```diff
@@ -0,0 +1,53 @@
...
+	return func(next http.Handler) http.Handler {
+		return http.HandlerFunc(func(w http.ResponseWriter, r *http.Request) {
+			t := StartTimer()
+			Logger.Log().Debugf("timer %p started", t)
+			rc := r.Clone(context.WithValue(r.Context(), timerContextKey, t))
+			next.ServeHTTP(w, rc)
+			t.Stop()
+			Logger.Log().Debugf("timer %p stopped (%.6fs)", t, t.Duration)
```
lyoshenka, line 28: if you put this in a deferred fn, will it still fire in case of a panic? just curious

lyoshenka (approved):
lgtm, with some minor nits

internal/metrics/middleware.go
andybeletsky: It will. Probably worth measuring panicked calls as well.

PR #291, opened 2020-10-27
Cache full user

andybeletsky:


internal/storage/conn.go
```diff
@@ -67,15 +65,6 @@ func (c *Connection) Connect() error {
 	c.logger.WithFields(logrus.Fields{"dsn": dsn}).Info("connecting to the DB")
 	var err error
 	var db *sqlx.DB
-	for i := 0; i < maxDBConnectAttempts; i++ {
```
andybeletsky, line 70: @tiger5226 something to note here: I figured in an orchestrated environment (k8s or even docker-compose) we want the container to just fail and let the orchestrator manage its restart. Otherwise from the supervising entity standpoint it's just hanging for a few dozen seconds for no good reason, not serving any requests and not responding to liveness probes.

PR #306, opened 2021-02-08
Update const.go

jessopb:
add collection_list to allowed methods


tzarebczan:
LGTM! @andybeletsky can you merge when there's less activity later on?

tzarebczan (approved):

PR #315, opened 2021-03-10
Add wait for it to the docker image

tiger5226:


andybeletsky:
Maybe add it to `docker-compose.yml` as well?

tiger5226:
Once it was in an image I was going to play with it and see how to leverage it in a docker-compose. So before I update the compose, it would be good to test it out. Can you tell me the commands to build this image locally....successfully? Was struggling with this a bit. Then I can figure out if its any different, like if `.sh` can or should be used, if we need netstat or wget etc.

andybeletsky:
This is how it's [built](https://github.com/lbryio/lbrytv/blob/f34a96fb0469e7a39c8042d11708c4629b29a497/.github/workflows/integrate.yml#L77):

```
go get github.com/gobuffalo/packr/v2/packr2@v2.7.1
goreleaser --rm-dist --snapshot
docker build . -t lbrytv/lbrytv-api:localdev
```

kauffj:
ping on stuck PR

PR #316, opened 2021-03-11
Tighten up CORS validation

andybeletsky:
Eligible domains will be set in production config file


api/routes.go
```diff
@@ -53,14 +58,22 @@ func InstallRoutes(r *mux.Router, sdkRouter *sdkrouter.Router) {
...
 	v2Router.HandleFunc("/status", status.GetStatusV2).Methods(http.MethodGet)
-	v2Router.HandleFunc("/status", proxy.HandleCORS).Methods(http.MethodOptions)
+	v2Router.HandleFunc("/status", emptyHandler).Methods(http.MethodOptions)
 }
 
 func defaultMiddlewares(rt *sdkrouter.Router, internalAPIHost string) mux.MiddlewareFunc {
 	authProvider := auth.NewIAPIProvider(rt, internalAPIHost)
 	memCache := cache.NewMemoryCache()
+	c := cors.New(cors.Options{
+		AllowedOrigins:   config.GetCORSDomains(),
```
tiger5226, line 68: I have some changes coming through now to allow wildcards here. This seems like a big deal because there are so many subdomains in use. Also Jeremy is having the ui being automated with subdomains. So it will be https://<branch>.odyse.com. This will require wildcard matching. Would not restrict without it.

kauffj:
ping on stuck PR

api/routes.go
andybeletsky: Wildcards matching is supported in `CORSDomains` parameter of the config file.

Issue #329, opened 2021-05-04
Modify file naming scheme to append random characters at end of file name


tzarebczan:
https://github.com/lbryio/lbrytv/blob/master/app/publish/publish.go#L222

Append the random characters to the end of the original file name instead of the front. 

This way when a user downloads the files, they will have the title /claim name and then the random bits at the end.

andybeletsky:
This has been solved with the help of @nheingit, original file names are now fully preserved.

PR #331, opened 2021-05-05
(#329) - changed Handler method createFile's naming scheme to have th…

nheingit:
…e random characters added to the end of the file name instead of the front.


nheingit:
Hey Andrey, this is the issue I believe you sent me in our email. It seemed like a very quick fix. Did I miss something here, or is this what you were looking for?

andybeletsky:
Unfortunately, this will simply yield the original filename without any random characters at all.

PR #335, opened 2021-05-06
First lbry pr

nheingit:
I think this should fix it.


andybeletsky:
@nheingit please note the tests are failing.

nheingit:
Sorry I forgot to squash the commits. This should fix both the TestUploadHandler, as well as Test_fetchFile.

There were some other tests failing on my machine, but I think that's due to the previously discussed problems I'm having with the test suite.

andybeletsky:
Thank you @nheingit, the tests are passing now. While reviewing it, I realised the solution originally proposed is not ideal, as instead of `xxx_1Mb.dat` users will have files named `1Mb.dat_xxx`. How about instead of simply adding random strings to the file name we create a randomly named folder and save the file under its original name in it before publishing? This would result in downloaded files being saved under the name uploader intended. 

Please let me know if you would like to go down further that path.

nheingit:
Yeah I can do that, but given the troubles I've had getting everything to play nicely on my machine it won't be until later this week.

nheingit:
Hey, I should be able to start on this tomorrow, I just wanted to make sure this is still what we want done?

Just create a Tempdir with the random bits and place the untouched file inside?

Do you want there to be any naming structure to the directory, or just the randomly generated string?

nheingit:
I'm sure the tests will be failing for this, but I can't yet run the tests on everything. Can you provide me the output of what this generates?

token_privkey.rsa
```diff
@@ -0,0 +1,39 @@
+-----BEGIN RSA PRIVATE KEY-----
```
andybeletsky, line 1: These keys should not be checked into the source control.

app/publish/publish.go
```diff
@@ -211,15 +211,23 @@ func (h Handler) saveFile(r *http.Request, userID int) (*os.File, error) {
...
 	if err != nil {
 		return nil, err
 	}
-	return ioutil.TempFile(path, fmt.Sprintf("%s_*", origFilename))
+	randomDir, err := ioutil.TempDir(uploadPath, "")
+	if err != nil {
+		return nil, err
+	}
+	finalFilePath := path.Join(randomDir, origFilename)
+	if err := ioutil.WriteFile(finalFilePath, []byte(origFilename), os.ModePerm); err != nil {
```
andybeletsky, line 227: This will just create a file containing it's own filename, why would we want that?

andybeletsky:
Github seems to be backed up at this time and the test run is queued. But I suspect you're right and it will indeed fail as test code hasn't been modified to accomodate the new logic. Can you run tests just for the `publish` package on your machine? From our conversation, I remember the tests you had failing were not related to this.

> Just create a Tempdir with the random bits and place the untouched file inside?

Yes, this was the idea.

Another issue stands out to me, there's no code to cleanup those temporary folders so they will stay forever on the filesystem, even though the file itself will be deleted after upload has been processed.

app/publish/publish.go
nheingit: I changed my key. I didn't notice that I added them, thanks for the catch.

Currently path output would look like  /upload_path/{user_id}/{random}/filename.ext`

I will change it to just be: /upload_path/{user_id}/{random}

I'll make an additional commit to add a cleanup.

Since I've been doing smaller projects I'm not as experienced in testing as I should be. I did a bit of digging after reading this and am indeed able to run the individual tests on my machine.

andybeletsky: We still need the original filename to be preserved. My comment was specifically about the `ioutil.WriteFile(finalFilePath, []byte(origFilename)` call, this will create a new file named after the original one and then write its own filename as a string into it. Why not just open a file and leave it empty?

nheingit: I think this is due to my own misunderstanding of the test that is being run on it, as the tests tell me that  there's no such file or directory if I don't write anything to the file and then try to open it.

Error:      	Received unexpected error:
            	            	open /tmp/20404/183730872/1Mb.dat: no such file or directory
            	Test:       	Test_fetchFile/http://127.0.0.1:38039/1Mb.dat

However if I DO write anything, it recognizes the file is there, but tells me its an invalid description.

Due to my lack of knowledge here I'm unable to determine if that's because the code I'm writing is incorrect, or if I'm unable to change the tests to pass the valid code that I wrote.

andybeletsky: > as the tests tell me that there's no such file or directory if I don't write anything to the file and then try to open it

Not easy to tell without seeing how you were trying to create that file.

Bad file descriptor error is due to the file being opened in read-only mode.

nheingit:
Pushed this for you to take a look at. All of the tests are passing, but couldn't think of a way to cleanup the tempfolder.

I don't know if that should be the responsibility of createFile? Open to ideas.

app/publish/publish.go
```diff
@@ -211,15 +211,26 @@ func (h Handler) saveFile(r *http.Request, userID int) (*os.File, error) {
...
 	if err != nil {
 		return nil, err
 	}
-	return ioutil.TempFile(path, fmt.Sprintf("*_%s", origFilename))
+	randomDir, err := ioutil.TempDir(uploadPath, "")
+	if err != nil {
+		return nil, err
+	}
+
+	file, err := os.Create(path.Join(randomDir, origFilename))
```
andybeletsky, line 227: Can simply return `os.Create` here.

andybeletsky:
There's already a cleanup procedure for the file itself in the `Handle` function, no reason why folder removal can't be done there as well.

nheingit:
After some poking around, and asking some of the gophers over in the golang discord I couldn't really find a way to obtain the path of `f.Name()` in the handler function. It seems that I'm breaking convention by not deleting `randomDir` in the function that I call it in. Obviously I can't do that if we're planning on adding something to the file afterwards.

Thoughts on how you might do this?

nheingit:
After further reflection it came to me. I'm not sure if there are any other files or directories that are stored under the /upload_path/{user_id} path.

If there isn't I believe my solution works.

andybeletsky:
We cannot assume our user is only ever going to be uploading a single file simultaneously so recursively deleting all files and folders in his directory doesn't seem like the way to go. 

`filepath.Dir(f.Name())` should get you the directory part of the path.

nheingit:
that was the first that I tried. f.Name only returns the name, and not the path.

Showcased here: https://play.golang.org/p/Iq9fM6tBuAW

andybeletsky:
In your showcase `f.Name()` returns `./filename`, relative to your current dir, and `filepath.Dir` correctly returns `.` because current dir is where you created it.

If you provide `os.Create` with a full path, `f.Name()` [will give it back to you](https://play.golang.org/p/eSrH13W2b0g).

nheingit:
Sorry I suppose I didn’t give context for that.

im having trouble getting access to the full path as in this example you have the Finalpath because you initialize the TempDir in that function.

my question would be how do you gain access to the path without that?

andybeletsky:
I don't understand what exactly is not working. What difference does it make that we do it in a different function?

nheingit:
> If you provide os.Create with a full path, `f.Name()` will give it back to you.

I was under the impression that in this context `f.Name()` was only returning the /file.ext portion of the path and not the entire thing. I must have not implemented it properly the first time I tried this, I apologize!

Thanks for your patience. All tests are passing and I made the changes as requested.

app/publish/publish.go
```diff
@@ -107,9 +108,12 @@ func (h Handler) Handle(w http.ResponseWriter, r *http.Request) {
 		op := metrics.StartOperation(opName, "remove_file")
 		defer op.End()
 
-		if err := os.Remove(f.Name()); err != nil {
-			monitor.ErrorToSentry(err, map[string]string{"file_path": f.Name()})
+		newFile, err := os.Create(f.Name())
```
andybeletsky, line 111: Why create the file here again? It's already created in `createFile`.

app/publish/publish.go
```diff
@@ -107,9 +108,12 @@ func (h Handler) Handle(w http.ResponseWriter, r *http.Request) {
...
+		if err != nil {
+			log.Error(err)
 		}
+		defer os.RemoveAll(filepath.Dir(newFile.Name()))
```
andybeletsky, line 115: It's necessary to log a potential error returned from `RemoveAll`, check how it's already being done in the file removal procedure that you deleted a few lines above.

nheingit: Fixed creating the file again. I also added back the logging that was being done previously.

andybeletsky: The cleanup block is already running deferred.

andybeletsky:
I think it's ready for merging after the extraneous comments and files are removed!

andybeletsky:
Sorry, I would appreciate it if you squashed git commit history, 16 commits seems a bit much! I'll merge it right away. Please send us your LBC wallet address so we can compensate you for the contribution.

nheingit:
YIKES! sorry about that.

just one moment.

Issue #330, opened 2021-05-04
Do not log params for sync_apply method


andybeletsky:
Currently, params for every JSON-RPC call are being logged for debugging purposes in `query.Caller.SendQuery`. Params for the `sync_apply` call should be excluded from logging.

jbshu4546:
PR created

PR #347, opened 2021-06-20
Issue #330: do not log params for sync_apply method

jbshu4546:
do not log params for sync_apply method


andybeletsky:
Thank you for your contribution @jbshu4546 but removing these params from error log isn't quite what we're after. We need to remove JSON-RPC params from debug logging for one particular method.

PR #348, opened 2021-06-21
Issue #330: don't log params for sync apply and update test

jbshu4546:
New PR here.  I wasn't sure if my test change was the correct one, but it does pass.  Please advise.

### DIFF
--- a/app/query/caller.go
+++ b/app/query/caller.go
@@ -255,10 +255,13 @@ func (c *Caller) SendQuery(q *Query) (*jsonrpc.RPCResponse, error) {
 
 	logFields := logrus.Fields{
 		"method":   q.Method(),
-		"params":   q.Params(),
 		"endpoint": c.endpoint,
 		"user_id":  c.userID,
 		"duration": c.Duration,
+	}
+	// Don't log query params for "sync_apply" method
+	if q.Method() != MethodSyncApply {
+		logFields["params"] = q.Params()
 	}
 	logEntry := logger.WithFields(logFields)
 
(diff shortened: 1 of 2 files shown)


app/query/caller.go
```diff
@@ -253,12 +253,16 @@ func (c *Caller) SendQuery(q *Query) (*jsonrpc.RPCResponse, error) {
...
-		"params":   q.Params(),
-		"endpoint": c.endpoint,
-		"user_id":  c.userID,
-		"duration": c.Duration,
+	logFields := logrus.Fields{}
+	// don't log params for "sync_apply" method
+	if q.Method() != MethodSyncApply {
+		logFields = logrus.Fields{
+			"method":   q.Method(),
+			"params":   q.Params(),
```
andybeletsky, line 261: We only want to omit _JSON-RPC query params_, not all params. It would not make much sense to omit method name or duration for sync apply calls, for example.

In other words, only `q.Params()` needs to be skipped for `sync_apply` log call.

jbshu4546: Understood, my apologies.  If you keep this open I can get a fix up today.

Thank you for your patience.

jbshu4546:
Pushed a change.

andybeletsky:
This seems to be doing the job. But to make sure, it would be great to have a test for it. I would also appreciate cleaner commit history.

jbshu4546:
Roger.  Test up next.

jbshu4546:
I opted to add asserts to the existing caller test.

andybeletsky:
Thank you @jbshu4546, it's deployed and running in production now. Please send us your LBC wallet address so we can compensate you for the contribution.

PR #356, opened 2021-08-04
Update openapi.yaml

mayeaux:
Making docs a bit more clear for future devs


andybeletsky:
This file is auto-generated from the design package so any modifications have to be done in the `design.go` file. I ported your changes, thank you

PR #358, opened 2021-08-08
Adding Oauth to Odysee APIs

tiger5226:


app/wallet/wallet.go
```diff
@@ -30,7 +31,8 @@ func DisableLogger() { logger.Disable() } // for testing
 
 // TokenHeader is the name of HTTP header which is supplied by client and should contain internal-api auth_token.
 const (
-	TokenHeader = "X-Lbry-Auth-Token"
+	TokenHeader         = "X-Lbry-Auth-Token"
```
andybeletsky, line 34: This should be renamed to `LegacyTokenHeader`.

cmd/db_migrate_down.go
```diff
@@ -14,6 +17,15 @@ var dbMigrateDown = &cobra.Command{
...
+		var nrMigrations int64
+		if len(args) > 0 {
+			var err error
+			nrMigrations, err = strconv.ParseInt(args[0], 10, 32)
+			if err != nil {
+				logrus.Error("non integer passed as argument to migration")
+			}
+		}
+
+		storage.Conn.MigrateDown(int(nrMigrations))
```
andybeletsky, line 29: 👍

app/wallet/oauth.go
```diff
@@ -0,0 +1,199 @@
...
+	for _, aud := range info.Aud {
+		if aud == oauthClientID {
+			audienceRespected = true
+		}
+	}
+	if !audienceRespected {
+		return errors.Err("this token is not meant for Odysee APIs")
+	}
+
+	/* If we could get the http request we could valid allowed sources
```
andybeletsky, line 142: What needs to happen for that? Also, looks like there is a typo.

app/wallet/remote.go
```diff
@@ -47,3 +49,35 @@ func getRemoteUser(url, token string, remoteIP string) (remoteUser, error) {
 
 	return ru, nil
 }
+
+func getRemoteUser(url string, token oauth2.TokenSource, remoteIP string) (remoteUser, error) {
+	op := metrics.StartOperation(opName, "get_remote_user")
+	defer op.End()
+
+	c := lbryinc.NewOauthClient(token, &lbryinc.ClientOpts{
```
andybeletsky, line 57: The two functions are identical save for this part, perhaps we could combine them somehow.

app/auth/middleware.go
```diff
@@ -33,11 +34,32 @@ func Middleware(provider Provider) mux.MiddlewareFunc {
 	}
 }
 
+// LegacyMiddleware tries to authenticate user using request header
+func LegacyMiddleware(provider Provider) mux.MiddlewareFunc {
```
andybeletsky, line 38: Not much difference between `LegacyMiddleware` and `Middleware` as well.

app/wallet/oauth.go
```diff
@@ -0,0 +1,199 @@
+package wallet
```
andybeletsky, line 1: We need tests for this code.

andybeletsky (changes requested):

andybeletsky:
Seems like it could work. However, it's hard to reason about it when there is no tests for this key piece of functionality. They can be against a real or a mock oauth server instance. Some existing tests call actual internal-apis.

app/wallet/oauth.go
tiger5226: Well it would be best if the middleware would just pass the http request to the provider instead of the token. However, it's designed this way, from what I saw to make testing easier. If you had to process a request in the provider, using mock providers would be more costly and I found them being used in tests all over the place. So I did not want to make that sort of change.

app/auth/middleware.go
tiger5226: Yeah, this was specific choice to minimize risk. I suspect the `LegacyMiddleware` will be completely removed in a few months, so to make sure it was a cleaner split I just created a new middleware and named the old one Legacy.

app/wallet/remote.go
tiger5226: Agreed, meant come back to this with a switch parameter or something, forgot. Will push a change up.

PR #360, opened 2021-08-16
make publish.fetchFile retry on failed download

### LINKED ISSUE #349: Make publish.fetchFile retry on failed download

quantonganh:
Fix #349

### DIFF
--- a/app/publish/publish.go
+++ b/app/publish/publish.go
@@ -228,6 +228,13 @@ func (h Handler) createFile(userID int, origFilename string) (*os.File, error) {
 	return os.Create(path.Join(randomDir, origFilename))
 }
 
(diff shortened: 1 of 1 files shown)


andybeletsky:
Thank you for your contribution, @quantonganh. I generally like the simple approach, however, this will not handle read timeouts and disconnects while reading the file, which can be quite large. 

Additionally, would it be possible to modify or add a test to cover the retry logic? No need to cover all failure scenarios but it would be nice to make sure it works on any of the common ones.

andybeletsky:
Closing this as it's been fixed in [another pull request](https://github.com/lbryio/odysee-api/pull/365), sorry.

PR #369, opened 2021-09-01
app/publish: improve error handling on fetch file

### LINKED ISSUE #368: look into EOF error / improve handling of remote url upload
Error seen by user: 
```
Error: "unexpected EOF: \"https://vods2.odysee.live/odysee-replay/4553807e23243572e2ab4829dc155ac91cc68c90_odysee_1630350117319.mp4\""

```

FIle: 
https://vods2.odysee.live/odysee-replay/4553807e23243572e2ab4829dc155ac91cc68c90_odysee_1630350117319.mp4

Request: 
```
remote_url: https://vods2.odysee.live/odysee-replay/acf7846f360e5f32b934f52e57ed24170bfe9508_odysee_1630005397.mp4
json_payload: {"jsonrpc":"2.0","method":"publish","params":{"name":"test22222222","title":"test live - remote","locations":[],"bid":"0.01000000","languages":["en"],"tags":[],"thumbnail_url":"https://spee.ch/0/a6ce1b684cb8f73d.png","blocking":true,"preview":false,"license":"None","release_time":1629733412},"id":1630425026258}
```

you'll need to pass an x-auth-token from odysee internal apis along with the request (see odysee.com calls for example).

rnd:
Before this we added a typed error on publish package to determine what
errors that should be retried and exclude request error like io.ErrUnexpectedEOF.

Problem is that error can happens on request parse too.

This commit fix the issue by separating request error and fetch error
into separate type.

Fixes #368

### DIFF
--- a/app/publish/errors.go
+++ b/app/publish/errors.go
@@ -1,0 +1,21 @@
+package publish
+
+import "fmt"
+
+// FetchError reports an error and the remote URL that caused it.
+type FetchError struct {
+	URL string
+	Err error
+}
+
+func (e *FetchError) Unwrap() error { return e.Err }
+func (e *FetchError) Error() string { return fmt.Sprintf("fetch error on %q: %s", e.URL, e.Err) }
+
+// RequestError reports an error that was due invalid request from client.
+type RequestError struct {
+	Err error
+	Msg string
+}
+
+func (e *RequestError) Unwrap() error { return e.Err }
+func (e *RequestError) Error() string { return fmt.Sprintf("request error: %s", e.Err) }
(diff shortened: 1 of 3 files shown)


app/publish/error.go
```diff
@@ -0,0 +1,21 @@
+package publish
+
```
andybeletsky, line 2: The filename should be `errors.go`.

andybeletsky (approved):

andybeletsky:
Otherwise lgtm 👍

app/publish/error.go
rnd: updated :)

PR #374, opened 2021-09-07
app/publish: support resumable file upload

### LINKED ISSUE #202: Improve publishing on lbrytv
Currently publish procedure just posts a multipart form which is prone to failure for big files and unstable connections. We need resumable uploads, one option is to implement [TUS protocol](https://tus.io/protocols/resumable-upload.html) (needs to be discussed further with the frontend team)

rnd:
This introduces 3 new endpoints that implements TUS protocol.

- POST	/api/v2/publish
- PATCH /api/v2/publish/{file_id}
- HEAD	/api/v2/publish/{file_id}

And one more endpoint to interact with lbrynet server after the upload
is completed.

- POST /api/v2/publis/{file_id}/notify

What changes from the old publishing implementation are:
   - Now the upload procedures will be handled exclusively by TUS
     handlers (PostFile, PatchFile, and HeadFile)
   - Upon upload completion, client need to call Notify endpoint to
     inform odysee to forward the publishing request to lbrynet server.
   - User authentication happens on preUploadCreateHook

Fix #202

### DIFF
--- a/api/routes.go
+++ b/api/routes.go
@@ -22,6 +22,8 @@ import (
 	"github.com/gorilla/mux"
 	"github.com/prometheus/client_golang/prometheus/promhttp"
 	"github.com/rs/cors"
+	"github.com/tus/tusd/pkg/filestore"
+	tusd "github.com/tus/tusd/pkg/handler"
 )
 
 var logger = monitor.NewModuleLogger("api")
@@ -31,7 +33,10 @@ func emptyHandler(_ http.ResponseWriter, _ *http.Request) {}
 
 // InstallRoutes sets up global API handlers
 func InstallRoutes(r *mux.Router, sdkRouter *sdkrouter.Router) {
-	upHandler := &publish.Handler{UploadPath: config.GetPublishSourceDir()}
+	uploadPath := config.GetPublishSourceDir()
+	authProvider := auth.NewIAPIProvider(sdkRouter, config.GetInternalAPIHost())
+
+	upHandler := &publish.Handler{UploadPath: uploadPath}
 	r.Use(methodTimer)
 
 	r.HandleFunc("/", func(w http.ResponseWriter, req *http.Request) {
@@ -40,9 +45,9 @@ func InstallRoutes(r *mux.Router, sdkRouter *sdkrouter.Router) {
 	r.HandleFunc("", emptyHandler)
 
 	v1Router := r.PathPrefix("/api/v1").Subrouter()
-	v1Router.Use(defaultMiddlewares(sdkRouter, config.GetInternalAPIHost()))
+	v1Router.Use(defaultMiddlewares(sdkRouter, authProvider))
 
-	v1Router.HandleFunc("/proxy", upHandler.Handle).MatcherFunc(upHandler.CanHandle)
+	v1Router.HandleFunc("/proxy", upHandler.Handle).MatcherFunc(publish.CanHandle)
 	v1Router.HandleFunc("/proxy", proxy.Handle).Methods(http.MethodPost)
 	v1Router.HandleFunc("/proxy", emptyHandler).Methods(http.MethodOptions)
 
@@ -56,13 +61,33 @@ func InstallRoutes(r *mux.Router, sdkRouter *sdkrouter.Router) {
 	internalRouter.Handle("/metrics", promhttp.Handler())
 
 	v2Router := r.PathPrefix("/api/v2").Subrouter()
-	v2Router.Use(defaultMiddlewares(sdkRouter, config.GetInternalAPIHost()))
+	v2Router.Use(defaultMiddlewares(sdkRouter, authProvider))
 	v2Router.HandleFunc("/status", status.GetStatusV2).Methods(http.MethodGet)
 	v2Router.HandleFunc("/status", emptyHandler).Methods(http.MethodOptions)
+
+	composer := tusd.NewStoreComposer()
+	store := filestore.FileStore{
+		Path: uploadPath,
+	}
+	store.UseIn(composer)
+	tusCfg := tusd.Config{
+		BasePath:      "/api/v2/publish/",
+		StoreComposer: composer,
+	}
+
+	tusHandler, err := publish.NewTusHandler(authProvider, tusCfg, uploadPath)
+	if err != nil {
+		logger.Log().WithError(err).Fatal(err)
+	}
+
+	tusRouter := v2Router.PathPrefix("/publish").Subrouter()
+	tusRouter.HandleFunc("/", tusHandler.PostFile).Methods(http.MethodPost)
+	tusRouter.HandleFunc("/{id}", tusHandler.HeadFile).Methods(http.MethodHead)
+	tusRouter.HandleFunc("/{id}", tusHandler.PatchFile).Methods(http.MethodPatch)
+	tusRouter.HandleFunc("/{id}/notify", tusHandler.Notify).Methods(http.MethodPost)
 }
 
-func defaultMiddlewares(rt *sdkrouter.Router, internalAPIHost string) mux.MiddlewareFunc {
-	authProvider := auth.NewIAPIProvider(rt, internalAPIHost)
+func defaultMiddlewares(rt *sdkrouter.Router, authProvider auth.Provider) mux.MiddlewareFunc {
 	memCache := cache.NewMemoryCache()
 	c := cors.New(cors.Options{
 		AllowedOrigins:   config.GetCORSDomains(),
--- a/app/publish/http.go
+++ b/app/publish/http.go
@@ -1,0 +1,115 @@
+package publish
+
+import (
+	"context"
+	"crypto/x509"
+	"fmt"
+	"net/http"
+	"net/url"
+	"regexp"
+	"time"
+
+	"github.com/lbryio/lbrytv/app/auth"
+	"github.com/lbryio/lbrytv/app/wallet"
+	"github.com/lbryio/lbrytv/internal/errors"
+	"github.com/lbryio/lbrytv/internal/ip"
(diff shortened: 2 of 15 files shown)


api/routes.go
```diff
@@ -46,6 +48,21 @@ func InstallRoutes(r *mux.Router, sdkRouter *sdkrouter.Router) {
 	v1Router.HandleFunc("/proxy", proxy.Handle).Methods(http.MethodPost)
 	v1Router.HandleFunc("/proxy", emptyHandler).Methods(http.MethodOptions)
 
+	tusHandler, err := publish.NewHandler(uploadPath)
+	if err != nil {
+		logger.Log().WithError(err).Fatal(err)
+	}
+
+	tusRouter := v1Router.PathPrefix("/files").Subrouter()
```
andybeletsky, line 56: This probably should live under v2, with old handler living in the v1 alongside.

rnd: gotcha

app/publish/tus.go
```diff
@@ -0,0 +1,222 @@
...
+	if authProvider == nil {
+		return nil, fmt.Errorf("auth provider cannot be nil")
+	}
+
+	// make sure target directory is exist
+	defaultUploadPath := "./uploads"
+	if uploadPath == "" {
+		uploadPath = defaultUploadPath
+	}
+	if _, err := os.Stat(uploadPath); err != nil {
```
andybeletsky, line 52: It's okay to just call `os.MkdirAll(uploadPath, os.ModePerm)` here, it already does those checks and will not error if it exists.

app/publish/tus.go
```diff
@@ -0,0 +1,222 @@
...
+
+	// initialize tusd handler
+	composer := tusd.NewStoreComposer()
+	store := filestore.FileStore{
+		Path: uploadPath,
+	}
+	store.UseIn(composer)
+	h.composer = composer
+
+	config := tusd.Config{
```
andybeletsky, line 70: Feels wrong to define this configuration in the middle of the function, perhaps there's a better place for it?

app/publish/tus_test.go
```diff
@@ -0,0 +1,429 @@
...
+	"fmt"
+	"io/ioutil"
+	"net/http"
+	"net/http/httptest"
+	"os"
+	"path/filepath"
+	"strings"
+	"testing"
+
+	"github.com/gorilla/mux"
```
andybeletsky, line 14: 3rd party imports should be at the end

app/publish/tus_test.go
```diff
@@ -0,0 +1,429 @@
...
+		t.Fatalf("failed to create a new request: %v", err)
+	}
+	r.Header.Set(wallet.TokenHeader, "uPldrToken")
+	r.Header.Set("Upload-Length", fmt.Sprintf("%d", len(testData)))
+
+	w := httptest.NewRecorder()
+	newTestMux(h, auth.Middleware(newTestProvider)).ServeHTTP(w, r)
+
+	resp := w.Result()
+	if resp.StatusCode != http.StatusCreated {
```
andybeletsky, line 107: `github.com/stretchr/testify/assert` or `require` works really well to get rid of those conditionals, we use it everywere in the test code.

app/publish/tus.go
```diff
@@ -0,0 +1,222 @@
...
+		log.WithError(err).Error("failed to get upload info")
+		w.Write(rpcerrors.NewInternalError(err).JSON())
+		observeFailure(metrics.GetDuration(r), metrics.FailureKindInternal)
+		return
+	}
+
+	// NOTE: don't use info.IsFinal as it's not reflect the upload
+	// completion at all
+	if info.Offset != info.Size { // upload is not yet completed
+		err := fmt.Errorf("upload is still on process")
```
andybeletsky, line 152: In process?

api/routes.go
```diff
@@ -56,13 +59,23 @@ func InstallRoutes(r *mux.Router, sdkRouter *sdkrouter.Router) {
...
 	v2Router.HandleFunc("/status", status.GetStatusV2).Methods(http.MethodGet)
 	v2Router.HandleFunc("/status", emptyHandler).Methods(http.MethodOptions)
+
+	publishHandler, err := publish.NewHandler(authProvider, uploadPath)
+	if err != nil {
+		logger.Log().WithError(err).Fatal(err)
+	}
+
+	publishRouter := v2Router.PathPrefix("/publish").Subrouter()
+	publishRouter.HandleFunc("/", publishHandler.PostFile).Methods(http.MethodPost)
```
andybeletsky, line 72: It seems to me that we can post files completely unathenticated, it should not be the case.

app/publish/tus.go
rnd: ok, i can pass this as a param and move it to `InstallRouter` since we need the base path there too

api/routes.go
rnd: it doesn't actually, its a sub router of `v2` that already apply the default middlewares

app/publish/tus_test.go
rnd: gotcha, i prefer to use standard library as much as possible, and i don't mind more verbosity on the tests actually
but yea happy to change so it follows the rest of the test code

rnd:
updated, PTAL @andybeletsky :)

app/publish/tus_test.go
andybeletsky: Agree with using the standard library but tests are normally already verbose, every bit that helps to make them easier to read is helpful.

app/publish/tus.go
```diff
@@ -0,0 +1,204 @@
...
+
+// NewHandler creates a new publish handler.
+func NewHandler(authProvider auth.Provider, cfg tusd.Config, uploadPath string) (*PublishHandler, error) {
+	h := &PublishHandler{}
+
+	if authProvider == nil {
+		return nil, fmt.Errorf("auth provider cannot be nil")
+	}
+
+	// make sure target directory is exist
```
andybeletsky, line 46: While I'm strongly in favor of documenting and commenting the code, I would argue this comment is unnecessary as it is fairly obvious what `os.MkdirAll` does. Same goes for `tusd.NewUnroutedHandler` call below.

app/publish/tus.go
```diff
@@ -0,0 +1,204 @@
...
+type PublishHandler struct {
+	*tusd.UnroutedHandler
+
+	logger       monitor.ModuleLogger
+	composer     *tusd.StoreComposer
+	authProvider auth.Provider
+}
+
+// NewHandler creates a new publish handler.
+func NewHandler(authProvider auth.Provider, cfg tusd.Config, uploadPath string) (*PublishHandler, error) {
```
andybeletsky, line 39: An unwritten convention is that `NewSomething` should return `Something`, not `SomeSomething`. But again, I think the handler should just be renamed.

app/publish/publish.go
```diff
@@ -2,18 +2,16 @@ package publish
...
 	"io"
 	"net/http"
-	"net/url"
 	"os"
 	"path"
 	"path/filepath"
-	"regexp"
 	"time"
 
+	"github.com/gorilla/mux"
```
andybeletsky, line 14: Imports order

app/publish/tus.go
```diff
@@ -0,0 +1,204 @@
...
+	tusd "github.com/tus/tusd/pkg/handler"
+	"github.com/ybbus/jsonrpc"
+)
+
+const module = "tus.publish"
+
+// PublishHandler handle media publishing on odysee-api, it implements TUS
+// specifications to support resumable file upload and extends the handler to
+// support fetching media from remote url.
+type PublishHandler struct {
```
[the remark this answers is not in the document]
rnd: renamed it to TusHandler

rnd:
@andybeletsky the last commit fix authorization issue that i mentioned on friday, ptal :)

app/publish/http.go
```diff
@@ -97,3 +105,40 @@ func retryPolicy(ctx context.Context, resp *http.Response, err error) (bool, err
 
 	return false, nil
 }
+
+// parseRequestIP returns the request IP address.
+func parseRequestIP(headers http.Header, remoteAddr string) string {
```
andybeletsky, line 110: Is there any way we can re-use existing mechanisms for IP detection and user auth?

app/publish/tus.go
```diff
@@ -200,3 +202,35 @@ func (h TusHandler) lockUpload(id string) (tusd.Lock, error) {
...
+		logrus.Fields{
+			"module":         module,
+			"method_handler": "preCreateHook",
+		},
+	)
+
+	r := hook.HTTPRequest
+	user, err := userFromRequest(h.authProvider, r.Header, r.RemoteAddr)
+	if err != nil {
+		log.WithError(err).Error("error authenticating user")
```
andybeletsky, line 227: `Error` is too high a level for this message and the one below. We still want it logged in production but it's not going to be something we can act upon. `Info` would be enough.

app/publish/tus.go
```diff
@@ -200,3 +202,35 @@ func (h TusHandler) lockUpload(id string) (tusd.Lock, error) {
...
+// handlers overwrite the context with context background to avoid context
+// cancellation, and so any attempt to read values from request context won't
+// work here, until they can safely pass request context to TUS handler we need
+// to decouple before and after middleware to TUS hook callback functions.
+//
+// see: https://github.com/tus/tusd/pull/342
+func (h *TusHandler) preCreateHook(hook tusd.HookEvent) error {
+	log := h.logger.WithFields(
+		logrus.Fields{
+			"module":         module,
```
andybeletsky, line 219: `h.logger` already has this field set up, there's no need to set it up again for every context logger

app/publish/http.go
rnd: updated to use the implementation from `ip` package

Issue #370, opened 2021-09-01
Do not log all JSON-RPC parameters


anbsky:
Call parameters are logged now like this:

```..."params":{"any_tags":[],"channel_ids":["deed21cf85c453e074b95949098fcd0c3dc65d69","9ab8d447066d328d8ebd115f725e701bd6fce93a","82f1d8c257d3e76b711a5cecd1e49bd3fa6a9de9","0871b29c7db6f1ec013694c8146f772539fcffa3","537305bc1bd71e2a5d76efacd607fd0f4d95da0c","c07ce3862335cfe9ba72054347c4ea7808ce4eaa","0e1aff187fbd3d2b2b16e5e8c213d3e782dafa98","a1303dc6a592605b08d60cac19ecbc1c877d7de7","2a294ea41312c6da8de5ebd0f18dbfb2963bb1a4","d6350f9158825662b99e4b5e0442bcc94d39bc11","3e7bb536a1b755b6eb64043ae3fa933d5935eaad","0518c93a80efeaf06fb4ac4d3fa72cc6390ac100","b8a2d928a3d3f44457a10abd0539b84e4a29c286","564789ffe2420980dcf9e00c4746feb5bc0daad0","7364ba2ac9090e468855ce9074bb50c306196f4c","e33372c0d8b2cdd3e12252962ee1671d66143075","3e63119a8503a6f666b0a736c8fdeb9e79d11eb4","1ce5ac7bab7f2e82af02305ced3c095e320b52e5","588ff6ffc72884e61e8d19a9412d6db9ec67167e", ...```

This is excessive and clogs up the logs. Would be good to limit the number of parameters logged to some reasonable number to avoid that.

lucasgpulcinelli:
I have created a pull request addressing this issue (#381), if something is not ideal please say it and I can work on it.

phormio:
This should be closed because #381 has been merged.

tzarebczan:
Thank you for the poke!

PR #381, opened 2021-10-04
(#370) Do not log all JSON-RPC parameters

lucasgpulcinelli:
I was able to solve this issue by creating a function to copy the map of parameters and cut lists inside it to only a fixed size, then this shallow copy is the one that gets logged. I hope this implementation is acceptable.

### DIFF
--- a/app/query/caller.go
+++ b/app/query/caller.go
@@ -259,9 +259,12 @@ func (c *Caller) SendQuery(q *Query) (*jsonrpc.RPCResponse, error) {
 		"user_id":  c.userID,
 		"duration": c.Duration,
 	}
-	// Don't log query params for "sync_apply" method
+	// Don't log query params for "sync_apply" method,
+	// and also log only some entries of lists to avoid clogging
 	if q.Method() != MethodSyncApply {
-		logFields["params"] = q.Params()
+		paramMap := q.ParamsAsMap()
+		paramCut := cutSublistsToSize(paramMap, maxListSizeLogged)
+		logFields["params"] = paramCut
 	}
 	logEntry := logger.WithFields(logFields)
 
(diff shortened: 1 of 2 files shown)


app/query/caller.go
```diff
@@ -362,3 +365,22 @@ func isErrWalletNotLoaded(r *jsonrpc.RPCResponse) bool {
 func isErrWalletAlreadyLoaded(r *jsonrpc.RPCResponse) bool {
 	return r.Error != nil && errors.Is(lbrynet.NewWalletError(0, errors.Err(r.Error.Message)), lbrynet.ErrWalletAlreadyLoaded)
 }
+
+// Makes a shallow copy of a map, cutting the size of the lists inside it
```
andybeletsky, line 369: Function comment should starts with function name [as per the convention](https://go.dev/blog/godoc).

app/query/caller.go
```diff
@@ -362,3 +365,22 @@ func isErrWalletNotLoaded(r *jsonrpc.RPCResponse) bool {
...
+// to at most num, made for declogging logs
+func cutSublistsToSize(m map[string]interface{}, num int) map[string]interface{} {
+	ret := make(map[string]interface{}, len(m))
```
andybeletsky, line 372: Specifying map size during initialization is not really necessary.

andybeletsky (commented):
Thank you for your contribution @lucasgpulcinelli, I think this is a good solution. However, on second thought, feels like it could be misleading if a dev tries to debug a situation and nothing indicates that more parameters are being sent. Would it be possible to add something like an ellipsis to the list of cut parameters? Like this:

```
"params":{"any_tags":[],"channel_ids":["deed21cf85c453e074b95949098fcd0c3dc65d69","9ab8d447066d328d8ebd115f725e701bd6fce93a","...99 skipped..."]
```

andybeletsky:
> possibly creating problems not related to the logging at all.

Good point, would it be possible to write a test to prove that it does not truncate params being passed to the actual JSON-RPC call?

PR #463, opened 2022-11-01
feat: support unlisted and scheduled content

tzarebczan:


app/query/processors.go
```diff
@@ -269,6 +277,12 @@ TagLoop:
 
 	switch accessType {
 	case accessTypePurchase, accessTypeRental:
+		var signErr error
+		if isLivestream {
+			signErr = errNeedSignedLivestreamUrl
+		} else {
+			signErr = errNeedSignedUrl
+		}
```
nikooo777, line 285:
```suggestion
		signErr := errNeedSignedUrl
		if isLivestream {
			signErr = errNeedSignedLivestreamUrl
		}
```

PR #484, opened 2023-04-26
Add tag patching for scheduled and unlisted content

anbsky:


app/query/processors.go
```diff
@@ -389,6 +405,57 @@ func resolve(ctx context.Context, c *Caller, q *Query, url string) (*ljsonrpc.Cl
...
 }
 
+// preflightHookClaimSearch patches tag parameters of RPC request to support scheduled and unlisted content.
+func preflightHookClaimSearch(_ *Caller, ctx context.Context) (*jsonrpc.RPCResponse, error) {
+	query := GetFromContext(ctx)
+	params := query.ParamsAsMap()
+	if hasSource, ok := params["has_source"].(bool); ok && hasSource {
+		notTags, _ := params["not_tags"].([]string)
+		anyTags, _ := params["any_tags"].([]string)
+		if !hasAnyTag(params["any_tags"], ClaimTagPrivate, ClaimTagUnlisted) {
```
tzarebczan, line 415: hasAnyTags takes 2 params, 3 are being passed here. So the hasAnyTag check doesn't seem to work.

tzarebczan: you can also just pass anyTags instead of later converting tags to string?

app/query/processors.go
```diff
@@ -389,6 +405,57 @@ func resolve(ctx context.Context, c *Caller, q *Query, url string) (*ljsonrpc.Cl
...
+			if !hasTag(notTags, ClaimTagPrivate) {
+				notTags = append(notTags, ClaimTagPrivate)
+			}
+			if !hasTag(notTags, ClaimTagUnlisted) {
+				notTags = append(notTags, ClaimTagUnlisted)
+			}
+			params["not_tags"] = notTags
+		}
+		if !hasTag(anyTags, ClaimTagScheduledShow) {
```
tzarebczan, line 424: something also isn't working properly here, it's still filtering future ones when passing the tag. This example had: 
image

I just adjusted the logic to also check for c:scheduled:hide, since we'll use that to show all future uploads. So check for either here.

app/query/processors.go
tzarebczan: so maybe the main issue is that it's not extracting anyTags and notTags correctly to begin with?

PR #487, opened 2023-05-09
feat: global mod bypass

tzarebczan:
see commit message

### DIFF
--- a/app/query/processors.go
+++ b/app/query/processors.go
@@ -325,6 +325,10 @@ TagLoop:
 		signErr = errNeedSignedLivestreamUrl
 	}
 
+	if isUserAMod(ctx, environ) {
+		return true, signErr
+	}
+	
 	if accessType == accessTypeUnlisted {
 		// check signature and signature_ts params, error if not present
 		signature, ok := params["signature"]
@@ -357,6 +361,7 @@ TagLoop:
 	if environ == iapi.EnvironTest {
 		iac = iac.Clone(iapi.WithEnvironment(iapi.EnvironTest))
 	}
+
 	switch accessType {
 	case accessTypePurchase, accessTypeRental:
 		resp := &iapi.CustomerListResponse{}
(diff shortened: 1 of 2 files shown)


pkg/iapi/responses.go
```diff
@@ -44,6 +44,43 @@ type CustomerListResponse struct {
 	} `json:"data"`
 }
 
+type UserMeResponse struct {
+	Success bool        `json:"success"`
+	Error   interface{} `json:"error"`
+	Data    struct {
+		Id                  int         `json:"id"`
```
anbsky, line 51: Initialisms must be capitalized if possible: `ID`, `ExperimentalUI`

pkg/iapi/responses.go
```diff
@@ -44,6 +44,43 @@ type CustomerListResponse struct {
...
+		Language            string      `json:"language"`
+		GivenName           interface{} `json:"given_name"`
```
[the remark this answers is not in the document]
tzarebczan: shouldn't have trusted the copy/paste, chatgpt doing a better job.

tzarebczan:
Gave it another shot, should be cleaner now. There's probably a way to even clean it up more, but should be fine.

app/query/processors.go
```diff
@@ -568,3 +573,26 @@ func getStatusResponse(_ *Caller, ctx context.Context) (*jsonrpc.RPCResponse, er
 	rpcResponse.Result = response
 	return rpcResponse, nil
 }
+
+// Check and weakly return if the user is a mod, if errors occur false is assumed
```
anbsky, line 577: A func’s doc comment should start with the function name.

anbsky (commented):
Would've preferred at least a test for user/me response as a canary for unexpected internal-api format changes but otherwise looks good.

tzarebczan:
Please ship when you're up if you can. We can add a test later, but the api should not change much.

